PR #60, opened 2013-11-20
Add in sharing of the next_db var via shm

jonludlam:
This is to support thin provisioning in whatever form it comes in.

Signed-off-by: Jon Ludlam jonathan.ludlam@eu.citrix.com


euanh:
@jonludlam: Is this still needed with thin-lvhd?    If so, could you please rebase and update the PR?

PR #62, opened 2013-12-05
Fix 'make clean'

germanop:
Source tree, once unpacked in the build directory is not
cleaned up with make clean.
This is fine for the build system, because it does the builds
in one shot, but during development it is useful to do builds
repeatedly while fixing errors discovered.

Signed-off-by: Germano Percossi <germano.percossi@citrix.com>


germanop:
@tmakatos, did you forget I already issued this PR? https://github.com/xapi-project/blktap/commit/0b12ebe9fe285ca870d03b7315a6a2b866a98a25

germanop:
Closed by 0b12ebe (even though not "properly")

PR #107, opened 2014-07-22
CA-140516: use /dev/zero as a source of a zeroed piece of memory

tmakatos:
Signed-off-by: Thanos Makatos thanos.makatos@citrix.com

### DIFF
--- a/drivers/block-vhd.c
+++ b/drivers/block-vhd.c
@@ -260,13 +260,13 @@ static void finish_data_transaction(struct vhd_state *, struct vhd_bitmap *);
 
 static struct vhd_state  *_vhd_master;
 static unsigned long      _vhd_zsize;
-static char              *_vhd_zeros;
+static char              *_vhd_zeros = NULL;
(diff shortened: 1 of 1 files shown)


jandryuk:
Hi, out of curiosity, why did this PR change from mmap MAP_ANONYMOUS to mmaping /dev/zero?  What is the benefit?  Thanks.

MarkSymsCtx:
From the historical comments on the internal ticket  (as the original developers are no longer here) it is more efficient as mmapping /dev/zero doesn't actually map pages out of ram whereas the MAP_ANONYMOUS did and running large numbers of tapdisk process resulted in consuming all of the Dom0 memory in the anonymous pages and thus reducing the system scalability.

jandryuk:
Thanks, @MarkSymsCtx .  We're moving OpenXT to a newer SELinux reference policy, and we need to add a rule allowing mmap /dev/zero.  This explains why the reasoning.  Thanks again.

PR #135, opened 2014-09-24
CA-147284: don't use non-async-signal-safe functions for SIGHUP

tmakatos:
This patch removes the usage of non-async-signal-safe functions when
handling SIGHUP and replaces it with an event.

### DIFF
--- a/drivers/tapdisk-server.c
+++ b/drivers/tapdisk-server.c
@@ -68,6 +68,8 @@ typedef struct tapdisk_server {
 		int                          backoff; /* exponential backoff
 							 factor */
 	} mem_state;
+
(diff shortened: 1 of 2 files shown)


drivers/tapdisk-server.c
```diff
@@ -321,6 +323,9 @@ tapdisk_server_close_tlog(void)
 static void
 tapdisk_server_close(void)
 {
+	if (likely(server.tlog_reopen_evid >= 0))
+		tapdisk_server_unregister_event(server.tlog_reopen_evid);
+
```
chandrikas: It might be a good idea to reset tlog_reopen_evid to -1 after unregistering the event

PR #138, opened 2014-10-03
CA-147893: don't die if hotplug-status gets arbitrarily written

tmakatos:
This might happen because something has gone horribly wrong (in udev and/or
XAPI).


tmakatos:
Closing PR as it is against xs64bit which is not closed and reopening against master (#154). Will create a merge ticket for port fix to Creedence.

PR #139, opened 2014-10-09
CA-148137: don't free NBD client if there are pending requests, plus unit tests

tmakatos:
This patch fixes a bug where we blindly deallocate the NBD client without
checking whether there are pending requests. We address this issue by
marking the client as "dead" and letting the last pending request to
complete to deallocate it.


drivers/tapdisk-nbdserver.h
```diff
@@ -98,5 +100,28 @@ int tapdisk_nbdserver_listen_unix(td_nbdserver_t *server);
 void tapdisk_nbdserver_free(td_nbdserver_t *);
 void tapdisk_nbdserver_pause(td_nbdserver_t *);
 int tapdisk_nbdserver_unpause(td_nbdserver_t *);
+void tapdisk_nbdserver_clientcb(event_id_t id, char mode, void *data);
+int tapdisk_nbdserver_reqs_init(td_nbdserver_client_t *client, int n_reqs);
+void tapdisk_nbdserver_free_client(td_nbdserver_client_t *client);
```
germanop: duplicated below

drivers/tapdisk-nbdserver.h
```diff
@@ -81,6 +81,8 @@ struct td_nbdserver_client {
 	struct list_head        clientlist;
 
 	int                     paused;
+
+	bool                    dead;
```
germanop: It compiles by chance, given stdbool.h is not explicitly included

drivers/tapdisk-nbdserver.c
```diff
@@ -200,9 +203,12 @@ tapdisk_nbdserver_free_client(td_nbdserver_client_t *client)
 	if (client->client_event_id >= 0)
 		tapdisk_nbdserver_disable_client(client);
 
-	list_del(&client->clientlist);
-	tapdisk_nbdserver_reqs_free(client);
-	free(client);
+	if (likely(!tapdisk_nbdserver_reqs_pending(client))) {
```
germanop: Sorry, I have to ask :-) : is it so unlikely to have pending requests?

drivers/tapdisk-nbdserver.h
```diff
@@ -98,5 +100,28 @@ int tapdisk_nbdserver_listen_unix(td_nbdserver_t *server);
 void tapdisk_nbdserver_free(td_nbdserver_t *);
 void tapdisk_nbdserver_pause(td_nbdserver_t *);
 int tapdisk_nbdserver_unpause(td_nbdserver_t *);
+void tapdisk_nbdserver_clientcb(event_id_t id, char mode, void *data);
+int tapdisk_nbdserver_reqs_init(td_nbdserver_client_t *client, int n_reqs);
+void tapdisk_nbdserver_free_client(td_nbdserver_client_t *client);
+td_nbdserver_client_t *tapdisk_nbdserver_alloc_client(td_nbdserver_t *server);
+bool tapdisk_nbdserver_contains_client(td_nbdserver_t *server,
```
germanop: Given you added a description for a trivial function like tapdisk_nbdserver_reqs_pending
why not adding one here? ;-)

drivers/tapdisk-nbdserver.h
```diff
@@ -98,5 +100,28 @@ int tapdisk_nbdserver_listen_unix(td_nbdserver_t *server);
 void tapdisk_nbdserver_free(td_nbdserver_t *);
 void tapdisk_nbdserver_pause(td_nbdserver_t *);
 int tapdisk_nbdserver_unpause(td_nbdserver_t *);
+void tapdisk_nbdserver_clientcb(event_id_t id, char mode, void *data);
```
germanop: It is not part of the fix nor the code re-factoring but.. given you are in documentation mode and this
function is used later on in the tests, do you mind writing a short doc? Pleeeeeeease :)

tests/test_nbdserver.c
```diff
@@ -0,0 +1,82 @@
...
+	tapdisk_server_unregister_event_Ignore();
+
+	memset(&server, 0, sizeof server);
+	INIT_LIST_HEAD(&server.clients);
+
+	client = tapdisk_nbdserver_alloc_client(&server);
+	assert(client);
+}
+
+void tearDown(void) {
```
germanop: Should not we be using this function to free what we allocated in setUp (alloc_client)?

germanop:
Very few minor issues. The fix is fine and the tests almost.
For the tests, I did not quite like the names but I could not come up with any suggestion. So, I shut up (oops, I said I do not like the names..).

Apart from this silly issues, this patch is going to be in master and not xs64bit.
Is that fine?

germanop:
I have just noticed that you changed tapdisk_nbdserver_free_client from static to a declaration
in the header file, in the first commit. That commit is to address the issue, while refactoring
happens in the second commit.

To be really a nitpicker, even the newly created function, tapdisk_nbdserver_reqs_pending,
can be static in the first commit and then re-factored in the second commit. But in this
case it is up to you.

It could be inlined, though.

drivers/tapdisk-nbdserver.c
tmakatos: Yes.

drivers/tapdisk-nbdserver.h
tmakatos: OK I'll include stdbool.h

drivers/tapdisk-nbdserver.h
tmakatos: OK.

drivers/tapdisk-nbdserver.h
tmakatos: Right.

drivers/tapdisk-nbdserver.h
tmakatos: OK I'll add one.

tests/test_nbdserver.c
tmakatos: We should, however the nature of the tests do not allow us to do so, e.g. if a test fails ```tearDown()``` doesn't know anything about the test itself and might end up doubly-freeing ```client```. We _could_ ensure that there are no memory leaks but I don't think it's worth it: they're only few of them with few bytes each.

tmakatos:
It's fine for the fix to go to master, though we should mark it as candidate fix for XS 6.5.

I'll take care of ```tapdisk_nbdserver_free_client```.

tmakatos:
I've added another commit addressing your comments. I haven't bothered with the two functions w.r.t. refactoring, it just too much rebasing required. If everything is OK I'll squash this new commit into the first one.

tests/test_nbdserver.c
germanop: Checking the pointer for NULL is not enough? I am expecting every test either not freeing it
or freeing it and setting the pointer to NULL.

tmakatos: No, it's not enough because a test can fail because of a test assertion so we won't have time to set the pointer to NULL.

germanop:
- [x] Remove duplicated declaration
- [x] Add stdbool
- [x] tapdisk_nbdserver_contains_client description
- [x] tapdisk_nbdserver_clientdb
- [ ] setUp memory leak
- [ ] Rebase

I think we can still do something about the memory leak. If not, at least put a comment in
the function so we know there is an issue.
About the rebase, squash your last commit I have the previous one saved so I can always compare.
Otherwsie I have to review the code twice, before and after the squash.

About the other "rebase" request, that's fine, I will do before pushing if you do not mind.

tests/test_nbdserver.c
germanop: So, how do know in tapdisk when a client structure is pointing to a valid location or not?
If tapdisk_nbdserver_free_client just frees the memory, how do we know if the pointer is
valid or not?

tmakatos: ```tapdisk_nbdserver_free_client``` may or may not free ```client```, depending on whether there are pending requests against it. After calling ```tapdisk_nbdserver_free_client``` one should not try to touch the pointer.

germanop:
The new version did not compile.
I fixed that and rearranged the code to have every commit really independent.
I sent you the new version by email.
Please, check and update the pull request.

Specifically:
- Moved code description of functions not related to the fix in the unit test (where they are used)
- Reinstated function order but added a declaration (first commit): the reason is that because of the
  functions swap, the diff was not clear in showing what the fix was.
- Removed the static keyword in function definition (commit 2) causing comiplation error

tmakatos:
Closing pull request as it is against xs64bit which is now closed, will reopen one for master.

PR #140, opened 2014-10-21
CA-147743: use blktap3 without kernel grant-copy

tmakatos:
This patch series allows blktap3 to be used without having to apply the kernel patch the export the grant-copy functionality to user space. This allows external users of blktap to use the tip instead of some very old version.

To disable grant-copy simply pass `--disable-gcopy` to `./configure`.

tapdisk uses grant-map+memcpy to move data around, which is probably slower than grant-copy, but no slower than blktap2.

Signed-off-by: Thanos Makatos thanos.makatos@citrix.com


MarkSymsCtx:
Grant map is now considered a problem in all cases, grant copy should be used and 4.x kernels have this in by default now..

PR #146, opened 2014-12-03
CA-148679: add CRC32 header to ring stats

tmakatos:
This allows for a lockfree production/consumption of I/O ring stats, as
the consumer can simply use the checksum to detect and ignore transient
inconsistencies.


germanop:
@tmakatos attached the fix for you patch as we have currently tested.
If it is fine for you, you can either add this on top of your pull request
(from Jorge Martin \<jorge.martin@citrix.com\>) or I can push directly the fixes version

[pull_146_fixed.txt](https://github.com/xapi-project/blktap/files/273060/pull_146_fixed.txt)

PR #147, opened 2014-12-05
CP-10439: move xen-ringwatch to blktap

tmakatos:
Move xen-ringwatch (from xen) and its cron job (from SM) to blktap, as this is the right place for it.

### DIFF
--- a/drivers/xen-ringwatch
+++ b/drivers/xen-ringwatch
@@ -1,0 +1,644 @@
+#!/usr/bin/python
+#
+# Copyright (C) 2011 Citrix Systems, Inc.
+#
+# This library is free software; you can redistribute it and/or modify
+# it under the terms of version 2.1 of the GNU Lesser General Public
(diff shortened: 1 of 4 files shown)


drivers/xen-ringwatch
```diff
@@ -0,0 +1,644 @@
...
+            except IOError:
+                # NB. racing unplug, any ring.write() may raise.
+                pass
+
+if __name__ == '__main__':
+    from sys import argv, stdout, stderr, exit
+    from getopt import gnu_getopt, GetoptError
+    from pprint import pprint
+
+    DEFAULT_PERIOD = 1 # secs
```
germanop: In the latest version of xen-ringwatch we had in the xen patch queue, this variable was set to 4.
This is the only difference.

Is it a mistake or on purpose?

MarkSymsCtx:
We've survived without this for a decade so closing.

PR #152, opened 2015-01-23
Cachable reads

jonludlam:


jonludlam:
Not for merging - this is just a placeholder to stash this implementation until a better one comes along :-)

Problems at the moment:
1. We require a VBD to be plugged to record the statistic. This means that the initial qemu reads go unreported. This _must_ be fixed.
2. The data are being reported via RRDD as a VBD metric, when it could be considered to be a VDI metric (?)
3. The data are being written in a very bespoke non-standard way.

PR #153, opened 2015-01-26
CA-148137: don't free NBD client if there are pending requests, plus unit tests

tmakatos:
This patch fixes a bug where we blindly deallocate the NBD client without
checking whether there are pending requests. We address this issue by
marking the client as "dead" and letting the last pending request to
complete to deallocate it.


tmakatos:
Review has already taken place on #139, we need a final ack for this one.

germanop:
Ack.
Are you pushing yourself or shall I do it?

PR #158, opened 2015-03-27
Update to support CentOS 7

rosslagerwall:


stefanopanella:
I have reviewed the change and it looks OK (stefano.panella@citrix.com)

PR #164, opened 2015-07-13
CA-161449: Stats between tapdisk and vdi traffic

martinjorge:
Added new stats structure to account tapdisk traffic. Requests issued
from tapdisk to the vdi will be found in /dev/shm/td3-<pid>/vdi-<minor>

Signed-off-by: Jorge Martin <jorge.martin@citrix.com>

### DIFF
--- a/drivers/tapdisk-blktap.c
+++ b/drivers/tapdisk-blktap.c
@@ -33,6 +33,7 @@
 #include "blktap.h"
 #include "tapdisk-vbd.h"
 #include "tapdisk-blktap.h"
+#include "tapdisk-metrics.h"
 #include "tapdisk-server.h"
 #include "linux-blktap.h"
 
@@ -67,6 +68,7 @@ struct td_blktap_req {
 	unsigned int            id;
 	char                    name[16];
 	struct td_iovec         iov[BLKTAP_SEGMENT_MAX];
+        struct timeval          ts;
 };
 
 td_blktap_req_t *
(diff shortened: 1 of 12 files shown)


stefanopanella:
Ok, thanks for the changes, it looks good to me, let's just see the impact of calling gettimeofdays() 4 times per request and if that is a problem, we can add a parameter to enable/disable the time stats.

franciozzy:
We already call gettimeofday() several times per request (I counted 7 on the past). This can easily be examined with "strace -c". When I looked at this before, I couldn't see such calls affecting performance. Ring0 also told me that they should be fairly lightweight.

Regardless of this particular topic, I believe an assessment by the performance team is imperative. They can easily test various hosts with and without the new metrics patches to see if the datapath performance has been affected.

Environments that are bottlenecked by CPU are preferred as they will show if we are using more or less CPU for the same kind of workload. Alternatively, we can observe CPU consumption during the tests (although this makes the whole thing less automated, as we currently don't collect CPU metrics for matching with specific workloads).

martinjorge:
Thanks for the message Felipe. I am running some XenRT tests now with and without the RPM installed, I am now waiting for the results. I will post here the results.

martinjorge:
zfims9yx

These are the results after running multiple tests, both with the RPM installed (yellow) and without (blue).
The X axis are the number of IOPS and the Y axis is the block size. There is a small overhead but they look very similar. Feel free to discuss this.

The graph can be found at http://perf/?t=4079

stefanopanella:
looks good to me :)

franciozzy:
I didn't check the environment the tests took place, but I'm happy these graphs are totally equivalent.

F.

martinjorge:
These were the parameters used:
- using tapdisk3
- 1 vbd on a Linux VM
- 1 thread
- 1 outstanding request
- random reads (to avoid read-ahead and merging side-effects)
- on an SSD SR
- on a range of block sizes (512-4M)

germanop:
Fix the first commit message.
A blank line is needed.

germanop:
Trailing line in the first commit message

germanop:
There is an improved version of this fix. Make sure this pull request is synchronized with the
one we are testing

germanop:
@martinjorge this PR is very out of sync with the one tested so far internally. Please, rectify

germanop:
Do not worry, I will push the rebased version of this patch.
Main differences:
- fixes spaces instead of tabs
- removed code relying on "err" because that variable was removed by the conflicting commit

PR #167, opened 2015-07-26
[RFC] Implement discard for AIO on certain file systems

robertbreker:
Discard support is announced to the guest, if each of the following conditions
is true:
* block-aio is used as the driver
* the image is stored on a filesystem that supports hole punching.
* the kernel and file system combination are whitelisted for supporting hole
  punching in tapdisk-util.c:is_hole_punching_supported_for_fd()
* tapback is not run with the --nodiscard overwrite

Discard requests are translated into the following request on the file system:
* fallocate(fd, (FALLOC_FL_KEEP_SIZE|FALLOC_FL_PUNCH_HOLE), offset, length)

This commit has been dev-tested using:
* v8 Windows PV drivers that include XenDisk and thereby implement discard
* Linux xen-blkfront that implements discard

Signed-off-by: Robert Breker <robert.breker@citrix.com>


robertbreker:
I've squeeze the changes and grouped them so they are easier to review.
The elephant in the room is still the potential request reordering of discards, as raised by Felipe.

robertbreker:
Regarding the reordering of requests:

robertbreker:
Regarding the reordering of requests:
* We don't need to worry about write requests that get moved behind discard. Discard is by definition an advisory instructions that doesn't necessarily need to have any effect according to the specification. (http://lxr.free-electrons.com/source/include/xen/interface/io/blkif.h#L74)
* We may need to worry about read requests that get moved behind discard.

As the next step I'd suggest:
- Confirm whether guests would issue "READ, DISCARD", or whether they would issue barriers "READ, BARRIER", DISCARD" or wait "READ, WAITFORREPLY, DISCARD" when issuing the described request order.
- In case we may need to handle "READ, DISCARD"- my current thinking is that we can get away with flushing the local tapdisk queue into the kernel queue using tapdisk_server_submit_tiocbs before calling fallocate64. 

This will need some more thoughts both from @malcolmcrossley and myself to be certain.

PR #173, opened 2016-01-26
Fix build against upstream Xen

andyhhp:
Ask for the compat API when building against the new userspace libraries.

The 'xc_evtchn_close()' call on the gnttab handle has always been broken; the
compiler can now spot it.

Signed-off-by: Andrew Cooper <andrew.cooper3@citrix.com>


andyhhp:
1) Because in Xen 4.7, libxc has been split in several different userspace libraries, to start providing a stable API/ABI.
2) By (failing) to compile blktap against upstream Xen
3) No

PR #177, opened 2016-03-17
CA-186157: Extract tapdisk stats to separate header

rafalmiel:
Exported tapdisk-metrics.h is impossible to include,
since it contains dependencies to other blktap headers
which for now remains private (tapdisk.h, tapdisk-utils.h).

This patch moves tapdisk stats struct to a header
that does not contains other dependencies,
so that it can be included by other components
that need access to tapdisk statistics structure

Signed-off-by: Rafal Mielniczuk <rafal.mielniczuk@citrix.com>

### DIFF
--- a/drivers/tapdisk-metrics-stats.h
+++ b/drivers/tapdisk-metrics-stats.h
@@ -1,0 +1,41 @@
+/*
+ * Copyright (C) Citrix Systems Inc.
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; version 2.1 only
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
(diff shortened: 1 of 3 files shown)


germanop:
Did anyone reviewed or acked this patch?

drivers/tapdisk-metrics-stats.h
```diff
@@ -0,0 +1,41 @@
...
+#ifndef TAPDISK_METRICS_STATS_H
+#define TAPDISK_METRICS_STATS_H
+
+#include <stdint.h>
+
+#define BT3_LOW_MEMORY_MODE 0x0000000000000001
+
+struct stats {
+    uint32_t version;
+    unsigned long long oo_reqs;
```
andyhhp: These must be altered to uint64_t before anyone starts using this header file.

drivers/tapdisk-metrics-stats.h
```diff
@@ -0,0 +1,41 @@
...
+
+
+#ifndef TAPDISK_METRICS_STATS_H
+#define TAPDISK_METRICS_STATS_H
+
+#include <stdint.h>
+
+#define BT3_LOW_MEMORY_MODE 0x0000000000000001
+
+struct stats {
```
andyhhp: "stats" is far too generic a name.  at the very minimum, it should be tapdisk_stats

germanop:
Agreed

germanop:
However, we should be using the same header file instead of keeping sync'ing them between different repos

PR #178, opened 2016-03-18
CA-205513: Removing bad fds from the fdset passed to select call.

pritha-srivastava:
SCTX-2269: Removing bad fds from the fdset passed to select call.

This will prevent the select call from failing because of a bad fd.

Signed-off-by: Pritha Srivastava <pritha.srivastava@citrix.com>

### DIFF
--- a/drivers/scheduler.c
+++ b/drivers/scheduler.c
@@ -24,6 +24,7 @@
 #include <unistd.h>
 #include <string.h>
 #include <sys/time.h>
+#include <fcntl.h>
 
 #include "debug.h"
 #include "tapdisk.h"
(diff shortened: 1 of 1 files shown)


chandrikas:
@pritha-srivastava : The changes to prevent a bad/closed fd from being added to the select set looks good. Should we also take this opportunity, if possible, to add logs that will help us understand how exactly a negative event ID is generated? This fix handles the symptom well but it would be nice to get to the bottom of the root cause.

pritha-srivastava:
@chandrikas: The following log
 EPRINTF("EBADF: Marking event dead, id: %d", event->id); 
in the code that I have added will help determine the cause of negative event id. In the code, the event id is returned from 3 places: 1) Due to EINVAL error, 2) Due to ENOMEM error 3) In all other cases This value of event id logged will help in determining what is the actual cause of the error.

chandrikas:
IIUC, in the cases of ENOMEM and EINVAL, we will never see your log because the event ID is not added to the event list. This leaves us with the third scenario and if you think printing the event ID is the first step to understanding the root cause, that is fine. Please add to the patchqueue after changing the ticket number in the commit message. If you want, you can add both - CA-205513 and SCTX-2269

germanop:
I am pushing the patch without the reference to the SCTX ticket but to the CA

PR #179, opened 2016-03-31
Poll blkif ring to reduce request latency

jjd27:
Service requests appearing in the blkif ring with lower latency by polling. This avoids the need to wait for an event-channel interrupt from the frontend to propagate to the backend.

Performance measurements show a ~25% performance improvement.

Excessive consumption of CPU is mitigated by only polling for a short time after the previous request has been seen in the ring, and by not polling at all if the backend domain's CPU utilisation is already above a certain threshold.


chandrikas:
Changes reviewed by @stefanopanella and @chandrikas

chandrikas:
Just noticed that internally we still refer to cumd (structure names, functions etc). Could you please rename those too @jjd27?

jjd27:
Where? `fgrep -r cumd *` returns no matches for me. Commit https://github.com/xapi-project/blktap/pull/179/commits/8c13ef9a3628ee728a692f1918400685d26cf37c removes all the mentions I could find. Can you be more specific, please?

chandrikas:
Sorry, I mean all the references to cum_* in cpumond.c and tapdisk-server.c

jjd27:
Oh, good point! I completely forgot about them. Additional commit coming up...

germanop:
This code does not apply cleanly any longer.

Moreover, although I do appreciate that for reviewer's convenience commits can be added,
the overall result cannot contain commits fixing other commits: if a commit (e.g., the one with
the bad daemon name) contains error, that commit needs fixing, not another commit.

If it is really hard to fix it because it is a very basic commit, we can make an exception.

germanop:
@jjd27 , I did it myself even though it should not have been up to me:
* I removed, the last and third-last commit and did those changes in the commits, where they
should have been in the first place (It took a while..).
* I also fixed another thing that was missed (see attached diff)
* Made the code apply on current master

You can find the code here https://github.com/germanop/blktap/tree/fix-pr-179
Please, let me know if you are find with this changes so I can finally push.

germanop:
@jjd27 I am gonna push my reviewed version

PR #180, opened 2016-04-15
CA-207760: Skip updating stats if ring is dead

chandrikas:
Tapdisk could have received a disconnect while IO is in flight. Don't
update stats in the callback if the ring is already dead.

Signed-off-by: Chandrika Srinivasan <chandrika.srinivasan@citrix.com>


MarkSymsCtx:
Change looks good.

PR #182, opened 2016-05-24
Change 'datasource type' of reported stats

kostaslamda:
Commit 793605d4357f4f229bf229ff5e45966f27e0ef93 under
xapi-project/rrd-transport matches the datasource type
name used internally by the RRD daemon with the name
datasources must report. In this case, 'absolute' must
be changed to 'gauge'

Signed-off-by: Kostas Ladopoulos Konstantinos.Ladopoulos@citrix.com

### DIFF
--- a/drivers/tapdisk-vbd.c
+++ b/drivers/tapdisk-vbd.c
@@ -1126,7 +1126,7 @@ tapdisk_vbd_produce_rrds(td_vbd_t *vbd) {
 			"\"description\": \"Number of I/O errors\",\n");
 	err += tapdisk_snprintf(buf, &off, &size, 3, "\"owner\": \"host\",\n");
 	err += tapdisk_snprintf(buf, &off, &size, 3,  "\"type\": "
-			"\"absolute\",\n");
+			"\"gauge\",\n");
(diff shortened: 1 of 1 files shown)


johnelse:
This looks fine to me.

germanop:
@johnelse would you leave a comment with your Reviewed-by tag as you want it to appear
in the commit message?

johnelse:
Sure

```
Reviewed-by: John Else <john.else@citrix.com>
```

edwintorok:
Comment in rrd.ml says:
```
    - Absolute: meaning that the incoming data is an absolute rate
    - Derive:   meaning that the rate must come from the difference between the
                incoming data and the previous value
    - Gauge:    meaning that the value isn't a rate at all (e.g. temperature, load avg)
```

If you are reporting the total number of I/O errors since tapdisk was running then Gauge is appropriate. If you are reporting the number of I/O errors in the past 5 seconds then Absolute is more appropriate (i.e. Absolute is something that resets its value every time it is read, see https://oss.oetiker.ch/rrdtool/doc/rrdcreate.en.html).

There is a conflict here that needs to be solved before merging though.

edwintorok (approved):

MarkSymsCtx:
This code has been removed, no longer required.

PR #185, opened 2016-07-18
CP-18192: Removed code for xlvhd thin-provisioning

kostaslamda:
Revert all changes related to thin provisioning.

Signed-off-by: Kostas Ladopoulos <Konstantinos.Ladopoulos@citrix.com>


germanop:
Strictly speaking, the code was not xenvm specific (apart from one commit), so I would find
more appropriate to say "thin provisioning" in the headline of the commit message.
In fact, the code has been committed with a message like
"Merged code for xlvhd thin-provisioning", so it would be wise to use a similar headline.

It would help to give a specific hint about the code being removed, too.
For example

> Removed all the code introduced with merge d921a00

Apart from this, the commit is OK.

PR #189, opened 2016-08-19
CA-217394: XSO-587 ERROR: errno -12 at tapdisk_control_accept

geosharath:
Prevent negative ids getting assigned to valid events in
tapdisk scheduler

Signed-off-by: Sharath Babu Sharath.Babu@citrix.com

### DIFF
--- a/drivers/scheduler.c
+++ b/drivers/scheduler.c
@@ -294,8 +294,8 @@ scheduler_register_event(scheduler_t *s, char mode, int fd,
 	event->id       = s->uuid++;
 	event->masked   = 0;
 
-	if (!s->uuid)
-		s->uuid++;
+	if (s->uuid < 0)
+		s->uuid = 1;
 
 	list_add_tail(&event->next, &s->events);
 


drivers/scheduler.c
```diff
@@ -294,8 +294,8 @@ scheduler_register_event(scheduler_t *s, char mode, int fd,
 	event->id       = s->uuid++;
 	event->masked   = 0;
 
-	if (!s->uuid)
-		s->uuid++;
+	if (s->uuid < 0)
+		s->uuid = 1;
```
chandrikas: Instead of checking for a negative uuid, maybe we could change uuid to an unsigned int? We don't need negative values anyway and we won't roll over as often.

geosharath: actually we need -ve values to signify an error condition and also there are other structs such as conn which use the same event ids. Incase if we were to change in one place and leave all others as int this will cause a valid positive event_id returned by scheduler_register_events to be treated as a -ve value since in other places its still int. On the other hand if we make all the event_ids as unsigned then there will be noway to propagate the error condition from one function to another. So I felt it was safe to loop around than play with the data types, the existing code also wanted to do the same thing except that the programmer assumed the value would loop around 0 by default.

swathicitrix: Here as per my initial understanding of code flow, the idle way would be,  assign unused ones if
we reach the max limit.

chandrikas: Fair point @geosharath, I didn't consider that event_id will have to be changed too. Looks good to me.

germanop:
This patch has not passed internal tests, what are we going to do?
I will remove the Reviewed and Patchqueue tag

PR #190, opened 2016-08-23
CA-219462: Remove lock.h and lock.c

kostaslamda:
The public functions provided by lock.h are not being invoked
in any place in the blktap code.

The standard blktap rpm (blktap-3.3.0.xs1098-xs7.0.1.x86_64.rpm at
the time of writing) provides the following shared objects:
- libblktapctl.so.0.1.1
- libvhdio-3.3.0.xs1098.so
- libvhd.so.0.1.1

The devel blktap rpm (blktap-devel-3.3.0.xs1098-xs7.0.1.x86_64.rpm)
provides the following static libraries:
- libblktapctl.a
- libvhd.a

None of the above exports symbols for the functions in lock.h.

*\* N.B. **
These functions were used in a utility called 'lock-util'.
'lock-util' stopped being built and packaged in commit 6449ea8
(moving from make to autotools and moving the spec file in a
closed repository).


germanop:
Digging a little bit more I found out that these functions were used for a utility called lock-util
(in fact there is a main in lock.c).
It was removed with commit 6449ea8b while moving to autotools and moving the spec file into a closed repository.
However, this transition brought with it the drop of lock-util without any comment or reason.

germanop:
Make sure the commit message reflects the fact that we are removing also an executable (lock-util) that has not been packaged for years (starting from commit 6449ea8).
So, in case we changed our mind, it would be easier to recover things

germanop:
Reviewed-by \<germano.percossi@citrix.com\>

PR #191, opened 2016-08-31
CA-220042: sigact.sa_mask is not initialized (CID:57552)

geosharath:

### DIFF
--- a/drivers/td-rated.c
+++ b/drivers/td-rated.c
@@ -1333,7 +1333,12 @@ rlb_siginit(void)
 		goto fail;
 	}
 
-	sigemptyset(&sigmask);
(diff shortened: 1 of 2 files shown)


MarkSymsCtx:
Looks good

MarkSymsCtx:
This exposed another suspect usage of sigemptyset, so we should fix that as well.

MarkSymsCtx:
Update looks good.

germanop:
The patch reviewed and the patch we have been testing do not match.

I wonder what we have been testing so far.
Once this is clarified let's remember to port the changes to the release branch

MarkSymsCtx:
Merged separately, so closing.

PR #192, opened 2016-08-31
CA-220041: blkif used after freed (CID:57556)

geosharath:


MarkSymsCtx:
looks good

PR #193, opened 2016-08-31
CA-220000: memory leak from variable path (CID:57593)

geosharath:


MarkSymsCtx:
Looks good.

PR #194, opened 2016-08-31
CA-220002: resource leak for memory allocated to pname (CID:57549)

geosharath:


MarkSymsCtx:
Looks good

PR #196, opened 2016-09-01
CA-215084: Bug fixes in lvm-util.c

kostaslamda:
If there were no lines to be consumed by the for loop, 'lvs' and 'pvs'
would be used uninitialized. When we break out of the loop, we check
that there has been at least 1 line consumed or we fail.

### DIFF
--- a/lvm/lvm-util.c
+++ b/lvm/lvm-util.c
@@ -132,10 +132,7 @@ lvm_open_vg(const char *vgname, struct vg *vg)
 		goto out;
 	}
 
-	for (;;) {
-		if (lvm_read_line(scan))
-			break;
-
+	for (i = 0; !lvm_read_line(scan); ++i) {
 		err = -EINVAL;
 		if (sscanf(line, _NAME" %"PRIu64" %d %d "_NAME" %"PRIu64, vg->name,
 			   &size, &lvs, &pvs, pvname, &pv_start) != 6) {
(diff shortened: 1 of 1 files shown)


lvm/lvm-util.c
```diff
@@ -122,67 +122,72 @@ lvm_open_vg(const char *vgname, struct vg *vg)
 	err = asprintf(&cmd, "vgs %s --noheadings --nosuffix --units=b "
 		       "--options=vg_name,vg_extent_size,lv_count,pv_count,"
 		       "pv_name,pe_start --unbuffered 2> /dev/null", vgname);
-	if (err == -1)
-		return -ENOMEM;
+	if (err == -1) {
```
germanop: This entire block of changes is unnecessary

germanop:
This commit is broken in principle: there are no segafaults to fix, hence most of the new goto label added (if not all of them) are unnecessary.

lvm/lvm-util.c
```diff
@@ -132,10 +132,7 @@ lvm_open_vg(const char *vgname, struct vg *vg)
 		goto out;
 	}
 
-	for (;;) {
-		if (lvm_read_line(scan))
-			break;
-
+	for (i = 0; !lvm_read_line(scan); ++i) {
```
germanop: In principle, this modification is not exactly the same as the code before: it does not add elegance because
the previous code have 2 break conditions, while here one break condition is moved into the for.

Moreover, the code as it is now, works anyway but it is just because there is always a newline at the end
of the output but if the output were a single line without newlines, it would break at the very end of the
loop without ever having the chance of increasing "i".

I think it is more reasonable to leave the **for** statement as is and add an **else** clause the the first **if**, like
```
if (lvm_read_line(scan))
    break;
else
    i = 1;
```

germanop:

PR #197, opened 2016-09-02
CP-18757: Refactor the way we use ctx->fd (CID: 28676)

geosharath:

### DIFF
--- a/vhd/lib/libvhd.c
+++ b/vhd/lib/libvhd.c
@@ -2652,7 +2652,7 @@ vhd_close(vhd_context_t *ctx)
 {
 	vhd_cache_unload(ctx);
 
-	if (ctx->file) {
+	if (ctx->fd != -1) {
 		fsync(ctx->fd);
 		close(ctx->fd);
 	}


germanop:
We are lacking context in the commit message.

However, after a closer look to the issue raised by coverity (it's not a bug, it is just coverity does
not understand it is fine), it seems to me that coverity is concerned about leaking a file descriptor.

The proposed change it is not gonna make coverity happier: if there were a fd leak (there aren't but
it thinks there are), assigning -1 to fd is just making the leak definitive (again, there aren't but let's
play this game)

geosharath:
[~germanop], yes this is not a bug so my change wont make any difference. Coverity knows that we are leaking fd since we are initializing it with open() and then it goes out of scope. However, if we re-initialize it using -1 at the time of close then coveirty would know that its not a fd that is being leaked but another value which is absolutely fine in going out of scope. Please let me know if my understanding is correct.

MarkSymsCtx:
We aren't leaking it, we are closing in the if block immediately above where you made the change. However Coverity doesn't have sufficient context to know that; as we check ctx->file  as the guard before closing it. Now, we know that in all cases if ctx->file is valid then so is fd and if fd is valid then so is ctx->file and thus it will be closed. However, coverity as an automated tool cannot know that the two values are so intimately linked hence the issue being raised. We should be using ctx->fd for all pieces of this so that Coverity can see that each and every time we enter vhd_close with a valid value in ctx->fd we will close that fd and thus not leak it.

geosharath:
@MarkSymsCtx yes thats true, coverity dosent have the context between ctx->file and ctx->fd hence it thinks we are leaking it even though we close the fd each time we enter vhd_close. I think replacing the if(ctx->file) to if(ctx->fd != -1) and then closing the fd would be more appropriate. Can you please share your thoughts on the same. My apologies if this is what you had suggested in the first place :)

vhd/lib/libvhd.c
```diff
@@ -2652,11 +2652,11 @@ vhd_close(vhd_context_t *ctx)
 {
 	vhd_cache_unload(ctx);
 
-	if (ctx->file) {
+	if (ctx->fd != -1) {
 		fsync(ctx->fd);
 		close(ctx->fd);
 	}
-
+
```
MarkSymsCtx: Spurious whitespace, remove.

MarkSymsCtx:
Looks good now.

PR #198, opened 2016-09-02
CA-220226: Fix bugs in part-util.c

kostaslamda:
- Remove overridden variable assignments
- Check that 'fd' is non-negative before calling 'close'
- Change 'sec_size' type from 'int' to 'unsigned int'


germanop:

PR #200, opened 2016-09-05
CA-220434: Return 'err' instead of 0

kostaslamda:
'_tap_ctl_list_tapdisk()' returned 0, regardless of success/failure.
'err' is saved throughout the function, so return it instead of 0.


germanop:
Code is fine but I do not agree with the commit message: there are still cases where the function returns an error.
However, once in the **do** loop, errors are handled with the **fail** label that, in turn, uses the **out** label that always returns 0.

Change the commit message accordingly: it is sufficient to say that some errors were treated as if they were success

germanop:

PR #202, opened 2016-09-05
CA-220442: Check length of socket name before copying

kostaslamda:
'struct sockaddr_un' member 'sun_path' is a fixed size, 108 byte array.
Check that 'name' is at maximum 107 characters long before copying.

### DIFF
--- a/control/tap-ctl-ipc.c
+++ b/control/tap-ctl-ipc.c
@@ -187,6 +187,13 @@ tap_ctl_connect(const char *name, int *sfd)
 
 	memset(&saddr, 0, sizeof(saddr));
 	saddr.sun_family = AF_UNIX;
+
+	if (unlikely(strlen(name) >= sizeof(saddr.sun_path))) {
+		EPRINTF("socket name too long: %s\n", name);
+		close(fd);
+		return -ENAMETOOLONG;
+	}
+
 	strcpy(saddr.sun_path, name);
 
 	err = connect(fd, (const struct sockaddr *)&saddr, sizeof(saddr));


control/tap-ctl-ipc.c
```diff
@@ -187,6 +187,13 @@ tap_ctl_connect(const char *name, int *sfd)
 
 	memset(&saddr, 0, sizeof(saddr));
 	saddr.sun_family = AF_UNIX;
+
+	if (unlikely(strlen(name) > sizeof(saddr.sun_path) - 1)) {
```
germanop: This line should be more readable by adding parenthesis around sizeof and -1 instead of relying on operator precedence.

I do see your point in raising an error instead of truncating and keep going

stefanopanella: I agree changing 'sizeof(saddr.sun_path) - 1" to '(sizeof(saddr.sun_path) - 1)' would make it a bit more readable

germanop: This is a random line taken from the linux kernel
```
if (unlikely(!bvl && (gfp_mask & __GFP_DIRECT_RECLAIM)))
```

I think whoever wrote the block layer knows C enough, nevertheless they felt the need to
make it readable.

There is a difference between readable and understanble

MarkSymsCtx: Or even pull the maximum allowed size out as a constant and use that, no parens required.

kostaslamda: apples and oranges...

germanop:
Don't quite understand the apple and oranges comment.
That said, there was no need to mention the 108 chars in the commit message, given we are
using sizeof.

I am getting the commit as is, I  think we spent far too much time on this one.

PR #203, opened 2016-09-05
CA-220444: Fix possible race condition

kostaslamda:
In function tap_ctl_make_device() in tap-ctl-allocate.c, try to unlink
without checking if the device is there first. If it does not exist, do
nothing. For all other errors, return 'errno'.


MarkSymsCtx:
Reviewed-By: Mark Syms <mark.syms@citrix.com>

PR #205, opened 2016-09-06
CA-220479: Remove nonsensical checks

kostaslamda:
Variable 'len' and 'message.u.info.length' are both of type 'size_t'.
It is impossible for 'len' to be negative.


germanop:
Code is fine.
Commit message about info.length unnecessary: for what concerns the comparison with '0', the fact that info.length cannot be negative is irrelevant, given 'len' itself cannot be negative.

PR #206, opened 2016-09-06
CA-220463: Ensure string fits in buffer before copying

kostaslamda:
'message.u.blkif.pool' is a fixed size char array. Make sure the C
string pointed to by 'pool' (with the terminateing '\0' char) fits
before copying it over.


germanop:

PR #207, opened 2016-09-06
CA-220481: Fixes in tap_ctl_stats_fwrite()

kostaslamda:
- Ensure that 'sysconf()' returns a positive value before assigning
  to 'bufz'. Otherwise, set 'err' and jump to 'out'
- In case 'mmap()' returns 'MAP_FAILED', do not change 'buf'
  to 'NULL' before jumping to 'out'; if 'buf != MAP_FAILED'
  'munmap(buf, bufsz)' is called

### DIFF
--- a/control/tap-ctl-stats.c
+++ b/control/tap-ctl-stats.c
@@ -105,11 +105,17 @@ tap_ctl_stats_fwrite(pid_t pid, int minor, FILE *stream)
 
 	prot  = PROT_READ|PROT_WRITE;
 	flags = MAP_ANONYMOUS|MAP_PRIVATE;
-	bufsz = sysconf(_SC_PAGE_SIZE);
+
+	err = sysconf(_SC_PAGE_SIZE);
+	if (err == -1) {
+		err = -errno;
+		goto out;
+	}
+
+	bufsz = err;
 
 	buf = mmap(NULL, bufsz, prot, flags, -1, 0);
 	if (buf == MAP_FAILED) {
-		buf = NULL;
 		err = -ENOMEM;
 		goto out;
 	}


germanop:
I am not sure about the sysconf fix.

Coverity is complaining because it knows sysconf can return a negative value.
We know that, in this case it, cannot happen.

With this fix, it will (probably) stop complaining about bufsz being negative (even though it cannot be,
given it is size_t) but it will start complaining about casting a negative to a size_t.

control/tap-ctl-stats.c
```diff
@@ -105,11 +105,17 @@ tap_ctl_stats_fwrite(pid_t pid, int minor, FILE *stream)
 
 	prot  = PROT_READ|PROT_WRITE;
 	flags = MAP_ANONYMOUS|MAP_PRIVATE;
-	bufsz = sysconf(_SC_PAGE_SIZE);
+
+	err = sysconf(_SC_PAGE_SIZE);
+	if (err < 1) {
```
germanop: Why <1 and not <0? AFAIK, errors are flagged with -1

germanop:
Apart from that, code is fine

control/tap-ctl-stats.c
kostaslamda: From the man page:
PAGESIZE - _SC_PAGESIZE
Size of a page in bytes. Must not be less than 1. (Some systems use PAGE_SIZE instead.)

germanop: sysconf signals errors with -1 and that it was we are looking for.
The fact that, for this specific value we cannot get 0 is just a happy coincidence. If you are so sure that it is impossible to get 0, even better, no need to bother.

Check the standard return code and that is -1.

PR #208, opened 2016-09-06
CA-220482: Check return value of fwrite()

kostaslamda:
'len' is unused, so do not save the return value of 'fwrite()';
instead, check that it was successful, and if not, set 'err'
appropriately.

### DIFF
--- a/control/tap-ctl-stats.c
+++ b/control/tap-ctl-stats.c
@@ -158,7 +158,10 @@ tap_ctl_stats_fwrite(pid_t pid, int minor, FILE *stream)
 			goto out;
 		}
 	}
-	len = fwrite("\n", 1, 1, stream);
+
+	if (fwrite("\n", 1, 1, stream) != 1) {
+		err = -EIO;
+	}
 
 out:
 	if (sfd >= 0)


germanop:
This fix makes the code semantically different.

In none of the previous versions this code was meant to fail if you could not write the newline.
I suggest to just log without returning an error.

Looking at the context of the code I found other interesting "behaviours" but I suggest to fix them in
a different pull request.

1. if select returns a negative number, errno is set: let's use that as a return value.
2. read return value is checked for non positive values and return in such case. The problem is that that value can just be -1 (not useful to return) or 0 (the loop would end cleanly in that case, no
need to return immediately without it). We should just just for negative values and return -errno.
3. fwrite does not set errno, so the fwrite handling in the loop could easily return 0 even if there is an error.

Bonus one
4. There is nothing in the loop taking care of interrupted read/writes

germanop:
Not sure about point 2, any more.
This function has many thing to improve. Let's leave 2 as is

germanop:
Sorry, today I am changing my mind continuously.

This function is used only from the CLI while invoking "tap-ctl stats".
Leave the function as is and let's return the error.

germanop:

PR #209, opened 2016-09-07
CA-220001: buffer not NULL terminated (CID:57557)

geosharath:
size of id is 40B, when msg length is > 39 it would cause
id not to be NULL terminated, hence we need to handle this
corner case.

Signed-off-by: <Sharath Babu sharath.babu@citrix.com>

### DIFF
--- a/drivers/block-nbd.c
+++ b/drivers/block-nbd.c
@@ -29,6 +29,7 @@
 #include <arpa/inet.h>
 #include <netinet/tcp.h>
 #include <netinet/in.h>
+#include <string.h>
 #include "tapdisk.h"
 #include "tapdisk-server.h"
 #include "tapdisk-driver.h"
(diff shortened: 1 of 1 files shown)


drivers/block-nbd.c
```diff
@@ -138,7 +139,17 @@ tdnbd_stash_passed_fd(int fd, char *msg, void *data)
...
-			sizeof(passed_fds[free_index].id));
+                       sizeof(passed_fds[free_index].id));
+
+	/*
+	 * Since size of id is 40B, if msg length is > 39 it would cause id
+	 * to overflow and not be null terminated. Hence, we need to handle 
+	 * this corner case.
+	 */
+
+	if(unlikely(strlen(msg) >= 40 ))
```
germanop: Writing explicit values is very fragile: consider sizeof or something else

germanop:
In other places we are taking a different approach when it comes to buffer overflows: instead of truncating the message we print errors and return.

Would you confirm that in this case truncating is fine?

geosharath:
this is not actually message, its the tapdisk driver image name. The only place I saw this being used is to get fd against a name in tdnbd_retreive_passed_fd() function. Here, we use strncmp so functionality wise it would not cause any problem (only first n bytes are compared). Please let me know if you feel this might break else where. However, I can think of a case where it might now work if we have first n bytes in the name as name and then they are different, is this case possible in the given environment ?

germanop:
If you think it is not going to break anything that's fine.

However, using 40 and 39 explicitly it's a little bit fragile

chandrikas:
I think truncating the msg is fine but agree with Germano. Let's use sizeof instead of explicit values.

MarkSymsCtx:
This breaks Storage motion, needs rework.

freddy77:
A buffer is NUL terminated, not NULL.

MarkSymsCtx:
Buffer is a fixed length is not required to be NULL terminated.

PR #210, opened 2016-09-07
CA-215114: Resource leak in vhd_util_coalesce_open_output()

kostaslamda:
There is a case where vhd_open() succeeds (0 is returned) but
the 'src' and 'dst' header block sizes are different.
'vhd_util_coalesce_open_output()' returns an error, but 'dst' is
not properly closed.

These 2 distinct cases (opening the vhd and checking if it is
correct) are now handled seperately.

### DIFF
--- a/vhd/lib/vhd-util-coalesce.c
+++ b/vhd/lib/vhd-util-coalesce.c
@@ -570,13 +570,27 @@ vhd_util_coalesce_open_output(vhd_context_t *dst,
 	}
 
 	err = vhd_open(dst, name, VHD_OPEN_RDWR | flags);
-	if (err || dst->header.block_size != src->header.block_size) {
-		printf("error opening %s: %d\n", name, (err ? : EINVAL));
-		unlink(name);
-		return err ? : EINVAL;
+	if (err) {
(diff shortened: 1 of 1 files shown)


vhd/lib/vhd-util-coalesce.c
```diff
@@ -570,13 +570,27 @@ vhd_util_coalesce_open_output(vhd_context_t *dst,
...
+		printf("error opening %s: %d\n", name, err);
+		goto fail1;
+	}
+
+	if (src->header.block_size != dst->header.block_size) {
+		printf("source and destination VHD block sizes are different: "
+				"%"PRIu32" != %"PRIu32,
+				src->header.block_size,
+				dst->header.block_size);
+		err = EINVAL;
```
germanop: This function returns negative values. Here you are setting a positive one

germanop:
Fix EINVAL

vhd/lib/vhd-util-coalesce.c
MarkSymsCtx: It was before though

PR #211, opened 2016-09-07
CA-220515: Return correct error val in vhd_util_pathcmp()

kostaslamda:
'canonpath()' does not set 'errno' in case of error, so it is
useless using it in 'vhd_util_pathcmp()'.

Return '-ENOENT' instead.

Signed-off-by: Kostas Ladopoulos <konstantinos.ladopoulos@citrix.com>

### DIFF
--- a/vhd/lib/vhd-util-coalesce.c
+++ b/vhd/lib/vhd-util-coalesce.c
@@ -258,13 +258,13 @@ vhd_util_pathcmp(const char *a, const char *b, int *cmp)
 
 	apath = canonpath(a, __apath);
 	if (!apath) {
-		err = -errno;
+		err = -ENOENT;
 		goto out;
 	}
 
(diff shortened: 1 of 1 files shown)


germanop:
canonpath can or cannot set errno depending on the point it returns.
In fallback behaviour, it calls realpath that does set it. Otherwise it just returns.

I'd rather fix canonpath by making sure it does set errno on failure but I think it fails only when calling realpath.

I don't think the fix is correct or achieving anything.

germanop:
If we agree this is a false positive, then we need to teach coverity how to realize that (or stop complaining)

kostaslamda:
Apparently Coverity is failing to understand much simpler stuff; have a look at the single new defect in the last snapshot. We can have a chat about this when you get back.

PR #214, opened 2016-09-08
CA-216250: Fix resource leak

kostaslamda:
In 'vhd_util_coalesce_load_chain()', if there is an error and we
jump to 'out', 'next' might not be freed.


MarkSymsCtx:
Looks good.

PR #215, opened 2016-09-08
CP-18038: Fix multiple resource leaks

kostaslamda:
In 'vhd_util_coalesce_load_chain()', cleaning up in case of an
error is wrongly implemented.

Basically, we try to figure out if the object pointed to by 'entry'
has made its way into the list or not. Although the rationale for
doing that is not wrong per se (checking if its list was empty,
which it technically is, until it is added to one), it needed
something extra to work; its 'list_head' member initialized by calling
'INIT_LIST_HEAD(&entry->next)'. This had to be done for every newly
allocated object.

The approach we took here is a bit different. Instead of doing the
above, which would have worked fine, we check if the object pointed
to by 'entry' and the last object in the list live in the same
address. If yes, do not do anything, 'vhd_util_coalesce_free_chain()'
will free it. If no, free 'entry'.

Also, ensure that 'entry->raw_fd' is not negative before calling
'close()' on it.

### DIFF
--- a/vhd/lib/vhd-util-coalesce.c
+++ b/vhd/lib/vhd-util-coalesce.c
@@ -402,10 +402,16 @@ done:
 	err = 0;
 out:
 	if (err) {
-		if (entry && list_empty(&entry->next)) {
+		/*
+		 * If 'entry->next.next' is NULL, it means 'entry'
+		 * is not part of list 'head'; hence, it won't be
+		 * freed by 'vhd_util_coalesce_free_chain()'
+		 * and needs to be freed explicitly.
+		 */
+		if (entry && !entry->next.next) {
 			if (entry->vhd.file)
 				vhd_close(&entry->vhd);
-			else if (entry->raw)
+			else if (entry->raw && (entry->raw_fd >= 0))
 				close(entry->raw_fd);
 			free(entry);
 		}


vhd/lib/vhd-util-coalesce.c
```diff
@@ -299,7 +299,7 @@ vhd_util_coalesce_load_chain(struct list_head *head,
 	char *next;
 	vhd_context_t *child;
 	int err, cmp, vhd_flags;
-	struct vhd_list_entry *entry;
+	struct vhd_list_entry *entry = 0;
```
germanop: What's the point given it is set to NULL 2 lines below?

germanop:
Given @MarkSymsCtx asked for a second opinion, I think the fix is overly complicated.
As far as the "entry" memory leak is concerned, splitting the **if** in 2 separate nested conditions is the easiest and more understandable thing to do.

Current code if condition is not straightforward without having a look at the previous lines in the function.
Moreover, even if someone agreed that the if condition makes sense, this condition makes the code
fragile: it is not by design that that condition is true and any later change would easily break the code.

My understanding of the code is
1. Check if entry is allocated: if yes, prepare a free operation
2. Check if entry contains a list that needs to be closed before deallocation.

Unfortunately, the 2 conditions are bound together and the failure of the second will make the code
skip free for the first one. That's why they should be split and nested.

If any, probably I would free entry and set it to null but it is the last time entry is used, so not strictly necessary.

kostaslamda:
The rationale of the code, as I understand it, is:
1) Check if `entry` is allocated
2) Check if `entry` is in `head`:
* if yes, do nothing; it will be `free`d by `vhd_util_coalesce_free_chain()`
* if no, `free` it here

Also, before freeing `entry`, we need to check if we opened a VHD or a RAW file and call the appropriate function

So we have to find a way to determine if entry is in `head` or not. There are 2 ways we can do that:

1) Check that `entry->next` is empty (existing approach):

This is a bit confusing, because `next` should actually be named `node` (it is of type `struct list_head` and has 2 members, `previous` and `next`). If `previous` and `next` point to `entry->next`, it means `entry` has not been added to any list and should be `free`d.

The reason it doesn't work is because `entry->next` is uninitialized and the check is always false.
`INIT_LIST_HEAD(&entry->next)` should be called immediately after every `calloc()` for this to work.

2) Check that entry and the pointer to the last entry in the list point to the same address (my fix):

The `list_last_entry(...)` macro returns a pointer to the last `struct vhd_list_entry` put in the list. If this pointer is the same as `entry`, then it is in the list.

I do not see another way that this could be done. Any suggestions?

germanop:
> The reason it doesn't work is because entry->next is uninitialized and the check is always false.
INIT_LIST_HEAD(&entry->next) should be called immediately after every calloc() for this to work.

The check *is not* always false.
`list_add(&entry->next, head);`
will correctly initialize the values but here we are discussing the case when that point is not reached.
So, *when* the check is false (because it is not always false).

INIT_LIST_HEAD is not needed to be called after `calloc`. `entry->next` is supposed to be an element
not a head. However, given it is not forbidden (just semantically wrong) it could be exploited (see later).

If entry is not added to any list, `entry->next` is a list node with both `next` and `prev` NULL.
So, if you want to know if entry is part of any list you just need to check if
- &entry->next->next is NULL
or.. given you suggested to use `INIT_LIST_HEAD(&entry->next)` after calloc
- you could just keep `list_empty(&entry->next)`

Now, the 2 methods above could be, according to my earlier reasoning, fragile because they rely
on the fact that if `entry` is part of a list, that list is necessarily `head` and we rely on this
when calling `vhd_util_coalesce_free_chain()`.
Instead your check is very accurate because it checks explicitly that entry is part of head and, if true,
you can safely rely on `vhd_util_coalesce_free_chain()` to free it.

The problem is that if we cannot rely on the fact that if entry is part of a list, that list has to be `head`,
then both methods are in trouble:
- yours because it would fail the check when entry is part of another list but not `head`, freeing `entry`
and causing a *segfault* when the other list is searched
- mine because would assume that `vhd_util_coalesce_free_chain()` would free it (when passed `head`
but it won't, causing *leaks*.


At last, the part where I do agree with you :-) 

Yes, splitting the if clauses and nesting is not right. Not sure if I had a good reason when
I said it last time and even if I had, I do not remember.

So many words to say that the core of your fix is right, it is just cumbersome and not very readable
the check if entry is in the list.

Given you asked for suggestions (plus I wanted to make clear that the if is not *always* false),
I verbosely gave a few :-)

germanop:

PR #216, opened 2016-09-09
CA-220808: Do not free unallocated pointers in vhd_close()

kostaslamda:
In the beginning of 'vhd_open()', 'ctx' is memset to 0; memory is
not allocated to all of the pointers in the object. When calling
'vhd_close()', check if all pointers are different from 0 before
freeing them.

Signed-off-by: Kostas Ladopoulos <konstantinos.ladopoulos@citrix.com>

### DIFF
--- a/vhd/lib/libvhd.c
+++ b/vhd/lib/libvhd.c
@@ -2668,12 +2668,21 @@ vhd_close(vhd_context_t *ctx)
 	if (ctx->file) {
 		fsync(ctx->fd);
 		close(ctx->fd);
+		free(ctx->file);
 	}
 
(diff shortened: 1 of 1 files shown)


andyhhp:
This patch is of no use at all. `free()` is explicitly specified to tolerate NULL pointers so you don't need to write code like this.  The old code as correct

kostaslamda:
It was flagged by Coverity.

andyhhp:
Coveritied complaint is that `free(ctx->custom_parent);` is uninitialised.  Sticking an extra `if()` doesn't make it any less uninitialised.

PR #217, opened 2016-09-12
CA-221814: Fix resource leak in vhdi_create()

kostaslamda:
In case file 'name' fails to open, memory allocated to
'buf' is leaked.


stefanopanella:
Reviewed-by: Stefano Panella

PR #218, opened 2016-09-12
CA-221864: Fix bugs in vhd_journal_read_locators()

kostaslamda:
- Return value from vhd_journal_position() not checked;
  ensure it is valid before continuing
- In case of error inside the for loop, always free 'buf';
  if 'NULL', memory has not been allocated (1st iter) or the
  pointer is already added to '_locators' (all other iters);
  if not 'NULL', pointer not copied to '_locators' yet


stefanopanella:
Reviewed-by: Stefano Panella

PR #219, opened 2016-09-13
CA-221904: vhd create EINVAL, report errors

MarkSymsCtx:
Signed-off-by: Mark Syms mark.syms@citrix.com

### DIFF
--- a/vhd/lib/libvhd.c
+++ b/vhd/lib/libvhd.c
@@ -3159,8 +3159,10 @@ __vhd_create(const char *name, const char *parent, uint64_t bytes, int type,
 
 	ctx.fd = open(name, O_WRONLY | O_CREAT |
 		      O_TRUNC | O_LARGEFILE | O_DIRECT, 0644);
-	if (ctx.fd == -1)
-		return -errno;
+	if (ctx.fd == -1) {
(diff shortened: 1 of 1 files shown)


MarkSymsCtx:
Fall out from the kernel build but if we fail we should report why.

vhd/lib/libvhd.c
```diff
@@ -3159,8 +3159,10 @@ __vhd_create(const char *name, const char *parent, uint64_t bytes, int type,
 
 	ctx.fd = open(name, O_WRONLY | O_CREAT |
 		      O_TRUNC | O_LARGEFILE | O_DIRECT, 0644);
-	if (ctx.fd == -1)
-		return -errno;
+	if (ctx.fd == -1) {
+        fprintf(stderr, "%s: failed to create: %d\n", name, -errno);
```
germanop: Why fprintf instead of VHDLOG?

MarkSymsCtx: Because, when run as td-util, VHDLOG is not setup and thus nothing comes out. All the code paths down through td-util use either printf or fprintf(stderr). Consequence of code being used by at least three different executables with different logging strategies. It could probably do with rationalising so that all the logging goes through a single point which then vectors the logging to the destination selected by the executable main function. That however is a much bigger task for another day.

PR #221, opened 2016-09-14
CA-216249: Check length of socket name before copying

kostaslamda:
Ensure that string fits in 'sun_path' before copying it over.


MarkSymsCtx:
Looks good.

PR #222, opened 2016-09-14
CA-222077: Fix bugs in td_fdreceiver_start()

kostaslamda:
- Change socket file descriptor type to 'int'
- Check return value of strdup(); fail if 'NULL'
- Check return value of 'snprintf()'; fail if 'path' is
  truncated or a negative value is returned
- In case of failure, free 'fdreceiver->path' first


stefanopanella:
Reviewed-by: Stefano Panella

PR #223, opened 2016-09-15
CA-222139: Resource leak in tap_ctl_allocate_device()

kostaslamda:
tap_ctl_allocate_device() will allocate memory to *devname, if it does
not already point to a command line argument. That memory needs to be
freed in both the success (in the calling function) and failure (in
the same function) cases.


stefanopanella:
I agree with @kostaslamda  that the resource leak triggere in `tap_cli_allocate(int argc, char **argv)` whe the -d option is not passed is not a false positive so it needs a similar fix as the one already present in this PR.

stefanopanella:
Reviewed-by: Stefano Panella

PR #224, opened 2016-09-16
CA-217896: tapdisk segfault

geosharath:
Tapdisk seg fault occured when a ring had pending
requests and is destroyed during a hard VM shutdown.
Those pending requests try to update the stats
which leads to a seg fault. Hence, gurading updates
to blkif->stats using blkif->dead during tapdisk
request operations.


swathicitrix:
Can you add a UT case for this ?
I mean try to simulate the issue to verify your fix.

  stats are being accessed from  tapdisk_xenblkif_parse_request() also. So code requires changes here also.

geosharath:
yes, I feel we need to add everywhere when we are accessing the stats, please share your thoughts on the same.

MarkSymsCtx:
Resolved in a separate change

PR #225, opened 2016-09-16
CA-222242: memset() 1 byte instead of 4

kostaslamda:
In 'vhd_kill_footer()', we call 'memset()' passing a 4 byte value
instead of a 1 byte value, as expected. Everything after the first
byte is truncated.


stefanopanella:
Reviewed-by: Stefano Panella

PR #226, opened 2016-09-16
CA-222244: Fix resource leaks in vhd_shift_metadata()

kostaslamda:
There is a case where 'buf' has memory allocated, the pointer
is not copied to locators[i] yet and we fail. 'buf' will be
leaked.

Fix:
- Initialize 'buf' to 'NULL'
- 'posix_memalign()' will set 'buf' to 'NULL' in case of error;
  remove redundant 'NULL'
- Set 'buf' to 'NULL' at the end of every iteration in the loop
- Free 'buf' in 'out'

In the second for loop, if 'vhd_check_for_clobber()' fails, all
allocated memory is leaked.

Fix:
Instead of returning -EINVAL, set it to 'err' and go to 'out'.

### DIFF
--- a/vhd/lib/vhd-util-resize.c
+++ b/vhd/lib/vhd-util-resize.c
@@ -689,7 +689,7 @@ vhd_shift_metadata(vhd_journal_t *journal, off64_t eob,
 	int i, n, err;
 	vhd_context_t *vhd;
 	size_t size_needed;
-	void *buf;
+	void *buf = NULL;
 	char **locators;
 	vhd_parent_locator_t *loc;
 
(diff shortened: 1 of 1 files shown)


vhd/lib/vhd-util-resize.c
```diff
@@ -729,6 +728,7 @@ vhd_shift_metadata(vhd_journal_t *journal, off64_t eob,
 			goto out;
 
 		locators[i] = buf;
+		buf = NULL;
```
stefanopanella: Maybe as optimization move this down out of the loop after the curly bracket.

stefanopanella:
Reviewed-by: Stefano Panella

PR #227, opened 2016-09-21
CA-217394: Handle scheduler uuid data type overflow in tapdisk

geosharath:
There are few events that persits throughout the life of
tapdisk process itself. Hence, when scheduler uuid overflow
occurs we should skip replacing these events for proper
functioning of tapdisk.

### DIFF
--- a/drivers/scheduler.c
+++ b/drivers/scheduler.c
@@ -261,6 +261,37 @@ scheduler_run_events(scheduler_t *s)
 }
 
 int
+scheduler_get_event_uuid(scheduler_t *s) {
+
+	int uuid_found;
+	event_t *event;
+
+        if(unlikely(s->uuid < 0)) {
+		EPRINTF("scheduler uuid overflow detected");
+                s->uuid = 1;
+                s->uuid_overflow = 1;
+        }
+
+        if(unlikely(s->uuid_overflow == 1)) {
+                do {
+                        uuid_found = 1;
+                        scheduler_for_each_event(s, event) {
+                                if(event->id == s->uuid) {
+                                        uuid_found = 0;
+                                        s->uuid++;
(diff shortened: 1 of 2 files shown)


jjd27:
This looks fine to me. But did you consider whether it be easier just to use an `unsigned long long int` for the ID rather than an `int`? A simple calculation shows that if we create events at the rate of 1000 per second, we'd need a tapdisk uptime of a billion years before overflowing.

geosharath:
@jjd27 @chandrikas changing the data type leads to cascading changes at various places in blktap. However, even if we change the datatype we are just delaying the occurrence of the issue and not exactly handling it. We got confirmation from customer that this patch is working fine in the mean time another customer is also reported facing this issue. Please let me know if you have any other concerns.

jjd27:
I presume the cascading changes are just changing the types of some variables? Or does it require something more complex? Personally I don't mind any problem that won't happen within a billion years. But it's not my decision which approach you should adopt. I was merely proposing what I assumed would be a simpler and less invasive change.

geosharath:
@jjd27 , yes its just changing the data type but we cant just change the data type without knowing the implications it might have on the entire system. event id is one such variable which gets accessed/processed by almost all the modules in blktap. Ensuring in all places where its accessed/processed that we have changed the data type and it does not have adverse implications might be tedious and error prone.

chandrikas:
@geosharath, I agree with your approach. This fix has also been validated by a customer. Changing the data type would also be ok but it will need more testing to be acceptable.

germanop:
May I ask to have the first line of the commit message a bit more descriptive of the fix and not just reporting the ticket name?
We do not even need 2 ticket references, the first CA is exhaustive.

PR #228, opened 2016-09-21
CA-222124: Handle race-condition in tap-ctl spawn

chandrikas:
When two tapdisk processes are spawned at the same time, they race
to create the /var/run/blktap-control directory. We currently don't handle
the error condition EEXIST where one tapdisk process finds that
the directory already exists.

Signed-off-by: Chandrika Srinivasan chandrika.srinivasan@citrix.com

### DIFF
--- a/drivers/tapdisk-control.c
+++ b/drivers/tapdisk-control.c
@@ -1452,11 +1452,15 @@ tapdisk_control_mkdir(const char *dir)
 			*ptr = '\0';
 
 		err = mkdir(name, 0755);
-		if (err && errno != EEXIST) {
-			err = -errno;
-			EPRINTF("failed to create directory %s: %d\n",
-				  name, err);
-			break;
+		if (err) {
+			if (errno != EEXIST) {
+				err = -errno;
+				EPRINTF("failed to create directory %s: %d\n",
+					name, err);
+				break;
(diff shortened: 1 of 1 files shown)


MarkSymsCtx:
Looks good to me but let's see what Marcus says as well.

mg12:
Looks good to me too.

chandrikas:
Thanks guys, will get this into the patchqueue then.

kostaslamda:
Mixed tabs and spaces!

chandrikas:
@kostaslamda : Conscious decision. Didn't want to introduce anymore tabs.

germanop:
blktap/tapdisk has been written explicitly with the linux kernel coding style: whitespaces are wrong, tabs are the right ones

MarkSymsCtx:
We need to make it consistent then as it's currently a mess of tabs and 4 and 8 space indentation.

germanop:
I am not against being consistent but adding spaces goes against it.
90% of the code is written with tabs, not my choice but this is the coding style.

germanop:
Not accepting this until the spaces are converted in tabs.
Regardless of coding conventions (that we have and are kernel-style) it is mixed within the same function.

PR #229, opened 2016-09-28
CA-223652: Add delay to reduce number of syslog messages

MarkSymsCtx:
Add delay in between attempts to pause the same vbd so that the syslog doesn't get hard spammed for no benefit

Signed-off-by: Mark Syms mark.syms@citrix.com

### DIFF
--- a/drivers/tapdisk-control.c
+++ b/drivers/tapdisk-control.c
@@ -926,6 +926,7 @@ static int
 tapdisk_control_pause_vbd(struct tapdisk_ctl_conn *conn,
 			  tapdisk_message_t *request, tapdisk_message_t * const response)
 {
+	struct timeval now, next = { 0, 0 }, interval = { 0, 10000 };
 	int err;
 	td_vbd_t *vbd;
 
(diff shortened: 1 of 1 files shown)


MarkSymsCtx:
The formatting of this function is a bit all over the place, could reformat the function as part of this change, maybe as a different commit in the set?

drivers/tapdisk-control.c
```diff
@@ -926,6 +926,7 @@ static int
 tapdisk_control_pause_vbd(struct tapdisk_ctl_conn *conn,
 			  tapdisk_message_t *request, tapdisk_message_t * const response)
 {
+	struct timeval now, next = { 0, 0 }, interval = { 0, 10000 };
```
stefanopanella: Should we do this 100ms ?

MarkSymsCtx: it currently seems happy at 10ms but it's easy enough to change

stefanopanella: ok, looks good to me than

stefanopanella:
Reviewed-by: Stefano Panella

stefanopanella:
Reviewed-by: Stefano Panella

PR #230, opened 2016-10-13
CA-225067: tap_ctl_{spawn,create}: move tapdisk to cgroup slice

jjd27:
The '-c' option allows the user to specify which cgroup slice the new tapdisk
process should be put in. If unspecified, it is moved to the default cgroup
slice.

This avoids the need to rely on the cgrulesengd daemon which fails to do its job
if the time it takes to write to the tasks file exceeds the lifespan of the
parent tapdisk (before it forks and exits). On a 4.4 kernel, it can take several
milliseconds to write to this file, so that kind of failure is very likely.

Signed-off-by: Jonathan Davies jonathan.davies@citrix.com

### DIFF
--- a/control/tap-ctl-spawn.c
+++ b/control/tap-ctl-spawn.c
@@ -172,8 +172,48 @@ tap_ctl_get_child_id(int readfd)
 	return id;
 }
 
+/* Move the process to the nominated cgroup slice
+ * Return 0 on success, -1 on error (with errno set) */
+static int
+tap_ctl_move_to_cgroup(int pid, const char *slice)
+{
+	FILE *f = NULL;
+	char *path;
+
+	if (!slice) {
+		errno = EINVAL;
+		return -1;
+	}
+
+	path = malloc(FILENAME_MAX);
+	if (!path) {
+		EPRINTF("malloc failed: %d\n", errno);
+		return -1;
+	}
+
+	if (snprintf(path, FILENAME_MAX, "/sys/fs/cgroup/cpu/%s/tasks", slice) >= FILENAME_MAX) {
+		EPRINTF("path truncated: %d\n", errno);
+		free(path);
+		return -1;
+	}
+
+	f = fopen(path, "we");
+	if (!f) {
+		EPRINTF("failed to write pid to cgroups task file '%s': %d\n",
+		    path, errno);
+		free(path);
+		return -1;
+	}
+
+	fprintf(f, "%d", pid);
+	fclose(f);
+
+	free(path);
+	return 0;
+}
+
 int
-tap_ctl_spawn(void)
+tap_ctl_spawn(const char *slice)
 {
 	pid_t child;
 	int err, id, readfd;
@@ -195,6 +235,18 @@ again:
 	id = tap_ctl_get_child_id(readfd);
 	if (id < 0)
 		EPRINTF("get_id failed, child %d err %d\n", child, errno);
+
+	/* Put the tapdisk in a cgroup slice (best-effort) */
+	if (!slice) {
+#ifndef TAP_CTL_NO_DEFAULT_CGROUP_SLICE
+		/* No option specified; move it to the default slice */
+		if (tap_ctl_move_to_cgroup(id, "/") < 0)
+			EPRINTF("failed to move tapdisk %d to default cgroup slice: %s; ignoring.\n", id, strerror(errno));
+#endif
+	} else {
+		if (tap_ctl_move_to_cgroup(id, slice) < 0)
+			EPRINTF("failed to move tapdisk %d to slice '%s': %s; ignoring.\n", id, slice, strerror(errno));
+	}
 
 	return id;
 }
--- a/control/tap-ctl-create.c
+++ b/control/tap-ctl-create.c
@@ -43,7 +43,7 @@
 
 int
 tap_ctl_create(const char *params, char **devname, int flags, int parent_minor,
-		char *secondary, int timeout)
+		char *secondary, int timeout, const char *slice)
 {
 	int err, id, minor;
 
@@ -51,7 +51,7 @@ tap_ctl_create(const char *params, char **devname, int flags, int parent_minor,
 	if (err)
 		return err;
 
-	id = tap_ctl_spawn();
+	id = tap_ctl_spawn(slice);
 	if (id < 0) {
 		err = id;
 		goto destroy;
--- a/control/tap-ctl.c
+++ b/control/tap-ctl.c
@@ -270,7 +270,8 @@ tap_cli_create_usage(FILE *stream)
 		"[-r turn on read caching into leaf node] [-2 <path> "
 		"use secondary image (in mirror mode if no -s)] [-s "
 		"fail over to the secondary image on ENOSPC] "
-		"[-t request timeout in seconds] [-D no O_DIRECT]\n");
+		"[-t request timeout in seconds] [-D no O_DIRECT] "
+		"[-c <cgroup-slice>]\n");
 }
 
 static int
@@ -278,6 +279,7 @@ tap_cli_create(int argc, char **argv)
 {
 	int c, err, flags, prt_minor, timeout;
 	char *args, *devname, *secondary;
+	char *slice = NULL;
 	char d_flag = 0;
 
 	args      = NULL;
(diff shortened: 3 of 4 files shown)


germanop:
This patch assumes cgroups is enabled in the kernel. It is true for XenServer, it is not necessarily true in general.

In that case I think your snprintf will fail.

jjd27:
Yes, it assumes cgroups are enabled, but does it on a best-effort basis. I think the snprintf will succeed but the subsequent fopen will fail. Hence it should fall into the `if (!f)` block which prints a warning but otherwise continues as before.

germanop:
Yep, you are right. I'll let you know as soon as I am done with the overall thing.

robertbreker:
This change looks good to me - am impressed that you went the extra mile to make the cgroup configurable.

control/tap-ctl-spawn.c
```diff
@@ -172,8 +174,42 @@ tap_ctl_get_child_id(int readfd)
 	return id;
 }
 
+/* Move the process to the nominated cgroup slice */
+static int
```
germanop: This function returns values that are never used.
I don't think we are ever using errno in the calling function to do anything meaningful, so
I'd suggest to just return -1 (errno value is already logged in EPRINTF) and let the caller decide
what to do.
See my comments when this function is called.

control/tap-ctl-spawn.c
```diff
@@ -196,5 +232,15 @@ tap_ctl_spawn(void)
 	if (id < 0)
 		EPRINTF("get_id failed, child %d err %d\n", child, errno);
 
+	/* Put the tapdisk in a cgroup slice (best-effort) */
+	if (!slice) {
+#ifdef TAP_CTL_USE_DEFAULT_CGROUP_SLICE
+		/* No option specified; move it to the default slice */
+		tap_ctl_move_to_cgroup(id, "/");
```
germanop: The comment here applies to the other invocation of this funciton as well.

Here we should check the return value and communicate how do we handle this situation.
So far, the invoked function has notified the error and passed the ball to the caller.
Here, according to your implementation, you decided to keep going without cgroups.
This decision needs to be communicated to the user.

In alternative, we can decide to abort.
This second scenario could be useful, for example, in case with a more recent kernel the new cgroups interface
will be used (I am not sure the old one will be kept in place).
In that case, if we silently continue we will spot the problem only when performance will degrade or worse.

control/tap-ctl-spawn.c
```diff
@@ -196,5 +232,15 @@ tap_ctl_spawn(void)
 	if (id < 0)
 		EPRINTF("get_id failed, child %d err %d\n", child, errno);
 
+	/* Put the tapdisk in a cgroup slice (best-effort) */
+	if (!slice) {
+#ifdef TAP_CTL_USE_DEFAULT_CGROUP_SLICE
```
germanop: I guess that, with this one, you want to make cgroup usage a compile-time switch. If it were a run-time option I would be more happy but
this is something that can be added later one. Just let me know if you think that it does not make sense to be
a run-time option.

However, you define it few line above, so I don't see how the \#ifdef code is not to be executed.

Please, clarify your intention.

germanop:
@jjd27 : the patch needs some modifications but I am OK to start testing it internally.

I need to modify it to make it compatible with #223 but it is really minor. Hope you don't mind.

control/tap-ctl-spawn.c
```diff
@@ -172,8 +174,42 @@ tap_ctl_get_child_id(int readfd)
...
+		return -EINVAL;
+
+	path = malloc(FILENAME_MAX);
+	if (!path) {
+		EPRINTF("malloc failed: %d\n", errno);
+		return -errno;
+	}
+
+	snprintf(path, FILENAME_MAX, "/sys/fs/cgroup/cpu/%s/tasks", slice);
+	path[FILENAME_MAX-1] = '\0';
```
germanop: Nope. It is always safe. It's guaranteed to write no more than size bytes, *including* null terminator

control/tap-ctl-spawn.c
jjd27: I agree with your point about setting `errno` and returning -1.

control/tap-ctl-spawn.c
jjd27: It's one of those things that all sensible implementations will enforce, but could cause headaches if this code was ever to be ported to other platforms. I always feel it's better to be defensive, but I'll happily remove it if you'd prefer.

control/tap-ctl-spawn.c
jjd27: Yes, the intention was for it to be a compile-time switch. So if some people don't like the "move to the default slice" being the default behaviour they can easily disable it in one place. I felt this was more appropriate for now than adding another command-line option, especially as there are no users that will use that at present.

control/tap-ctl-spawn.c
MarkSymsCtx: Actually, in this case it's entirely moot as if it truncates we will do the wrong thing. So, we should be checking the return value and if >= FILENAME_MAX then we should be erroring and not blindly carrying on. Pretty unlikely to happen tbh.

germanop: @jjd27 I am not aware of any snprintf implementation that does not add a null char at the end, apart from Window's _snprintf that is spelled differently, anyway.
I just don't want to add coverity noise to our already noisy report. If you are aware of other
snprintf implementations that don't do that, please let me know and I will be more than heppy to reconsider.

@MarkSymsCtx FILENAME_MAX is supposed to be large enough to contain the string even though the implementation does not say if it is including null or not.
However, given any other macro like that (e.g. PATH_MAX) states explicitly that it must include
null termination, I would say that it is the same for FILENAME_MAX.
This means that if we do truncate, yes, fopen will fail because the file is not what we meant but
it would have failed anyway, given the file we passed cannot exist on the system because it would violate the max lenght, anyway.

control/tap-ctl-spawn.c
germanop: OK, now I know what you meant. I am OK to have it a compile time option (but in a near future
I'd rather have it a run-time one) but in this case the user needs to modify the source code, while
a middle ground approach would be to let people decide what to do while compiling.
So, what about this:
* call the macro "TAP_CTL_NO_CGROUP"
* remove the define directive
* change the ifdef to
```
#ifndef TAP_CTL_NO_CGROUP
```
Doing this, the default behaviour is the same as you pictured while giving the user the option to
pass -DTAP_CTL_NO_CGROUP and disable it.

control/tap-ctl-spawn.c
germanop: I know that your intention was that but:
* I'd rather have a message saying "could not do that so I am doing this"
* I am worried that with the new interface coming that sysfs entry won't be there any longer.
This means that we will stop using (silently or just visible in logs when the above point will be addressed) cgroups and we will spot it in specific tests.

So, given using cgroups it is an important thing for us (otherwise you would not have spent
time on this :-D ), we cannot afford the luxury to have it ignored and be unnoticed.

control/tap-ctl-spawn.c
MarkSymsCtx: I did say it was unlikely to happen, but this is taking input from the user, so there is nothing (that I can see anyway) to stop the user supplying something that causes this to overflow.

control/tap-ctl-spawn.c
germanop: @jjd27 I was not thinking of tcp_ctl_close because nothing has been necessarily allocated and attached. I thought that killing the process would have been sufficient but it well be my analysis is very simplistic.

Let's not fail, then, and I will raise another ticket to fix it for later versions (I am seriously worried about
"silent" failures).

Would you just use the return code (I think we agreed -1 is sufficient) to notify the user about errors for different
cases?

Case:
1. blktap compiled without the MACRO defined (see comment about macro): we do not even try cgroups so
nothing to log
2. blktap compiled with the MACRO defined, no slice specified: we try to use the default slice but it fails (let's say
cgroups is not enabled in the kernel), so we log that we are not using cgroups at all
3. blktap compiled with the MACRO defined, slice specified: we log that we could not write in that slice and we are not using cgroups (or, given we are handling errors now, you can decide to try with default, your choice).

My point here is: **tap_ctl_move_to_cgroup** can just log the fact that it fails but it does not know how we are
going to handle that. That's why we need to check the return code and log accordingly in the caller.

germanop:
To summarize:

- [ ] no need for null terminate after snprintf (unless there are versions of it that don't handle that)
- [ ] The default cgroups thing needs to be disabled at compile time (let me know if my proposal is not clear)
- [ ] Check return code from **tap_ctl_move_to_cgroup** and log accordingly

germanop:
ping

jjd27:
I've posted a new version of the patch. Changes since v1:
 - make tap_ctl_move_to_cgroup return -1 on error with errno set (rather than returning -errno)
 - free path on snprintf error
 - don't NUL-terminate after snprintf
 - only move if TAP_CTL_NO_DEFAULT_CGROUP_SLICE not defined (to allow this to be disabled at compile time)
 - use EPRINTF to report failures of tap_ctl_move_to_cgroup (but continue regardless)

PR #234, opened 2016-11-23
CA-233266: The VM didn't give disk performance data

swathicitrix:


swathicitrix:
On Wrong location, i pulled my commit. So closing it. 
Sorry for the mess up here.

PR #239, opened 2017-05-10
Fix udev rules to prevent td devices being bound by Dom0 services

MarkSymsCtx:
Remove obsolete udev rules which no longer work.
Patch override copy of 65-md-incremental.rules to prevent Dom0 mdadm from binding the devices.


rosslagerwall:
Is mk/*spec.in still used?

Surely the spec file in the patchqueue needs to be updated?

MarkSymsCtx:
It is as well, the mk.*spec.in is used in XS7 and also it exists here as a canonical recipe for packaging external to internal XenServer build systems.

rosslagerwall:
LGTM

germanop:
I see 2 problems here (plus a bonus one, not strictly related with the intentions of this patch):
1. The patch in %triggerin could fail as soon as mdadm decides to change the comment at the
  beginning of their rule
2. To be in par with previous behaviour (unless there is a strong reason for not doing that), we should
  reinstate the old rule upon blktap uninstall.
3. (Bonus one but just for a future discussion) I do not think blktap is the right place for this rule

PR #240, opened 2017-05-17
Fix udev rules to prevent td devices being bound by Dom0 services

MarkSymsCtx:
Remove obsolete udev rules which no longer work.
Patch override copy of 65-md-incremental.rules to prevent Dom0 mdadm from binding the devices.


MarkSymsCtx:
Sorry, couldn't reopen 239 as Github wouldn't let me after I updated the branch :(

https://github.com/xapi-project/blktap/pull/239

PR #253, opened 2018-06-18
make O_DIRECT optional

nraynaud:
- retry open() with O_DSYNC if an EINVAL error was encountered

ZFS doesn't support O_DIRECT, so I adopted what I hope is the least intrusive changes to the codebase.

### DIFF
--- a/vhd/lib/libvhd.c
+++ b/vhd/lib/libvhd.c
@@ -44,6 +44,7 @@
 #include <libgen.h>
 #include <iconv.h>
 #include <limits.h>
+#include <stdarg.h>
 #include <sys/mman.h>
 #include <sys/stat.h>
 #include <sys/types.h>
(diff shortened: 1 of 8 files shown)


nraynaud:
Hello, 
can you explain your process and what it would take to merge this change please?
I just signed off my commit.

Thanks, 

Nicolas.

BobBall:
It'd be cleaner to refactor the open calls to a new method which will try first with the flags passed, then remove O_DIRECT and replace with O_DSYNC in the failure case.  Can we also get some logging when we switch from the default flags so we can see what's going on?

nraynaud:
@BobBall sure, can you suggest a name and a residing file for the new function?

nraynaud:
@BobBall I did my best to comply. 

What's next in the quest to get this merged?

vhd/lib/libvhd.c
```diff
@@ -117,6 +118,24 @@ old_clear_bit(volatile char *addr, int nr)
 	((uint32_t *)addr)[nr >> 5] &= ~(1 << (nr & 31));
 }
 
+int
+open_optional_odirect(const char *pathname, int flags, ...)
+{
+	mode_t mode;
+	if (flags & O_CREAT) {
```
BobBall: I wonder if this complexity is worth it - it'd be easier to make the wrapper always take a mode_t (which could be passed as -1 in the calls which are not O_CREAT) - although if Mark is happy then I have no objections.

nraynaud:
I have passed 0, it feels like the safest value.

BobBall:
Thanks - I'm happy.

PR #254, opened 2018-07-06
CP-24318: Closed resource leak in vhd_macx_decode_location()

eliascalocanecitrix:
Signed-off-by: Elias Calocane <elias.calocane@citrix.com>

### DIFF
--- a/vhd/lib/libvhd.c
+++ b/vhd/lib/libvhd.c
@@ -1507,13 +1507,15 @@ vhd_macx_decode_location(char *in, char *out, int len)
 
 	name = out;
 	ibl  = obl = len;
-
-	cd = iconv_open("ASCII", "UTF-8");
-	if (cd == (iconv_t)-1) 
+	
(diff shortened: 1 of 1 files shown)


eliascalocanecitrix:
@MarkSymsCtx I've amended the config now so hopefully it's okay

vhd/lib/libvhd.c
```diff
@@ -1509,11 +1509,15 @@ vhd_macx_decode_location(char *in, char *out, int len)
 	ibl  = obl = len;
 
 	cd = iconv_open("ASCII", "UTF-8");
-	if (cd == (iconv_t)-1) 
+	if (cd == (iconv_t)-1) {
+		iconv_close(cd);
```
MarkSymsCtx: This isn't required and in fact will fail. Only the chunk below is necessary. If `iconv_open` open fails then cd does not refer to an open conversion descriptor.

eliascalocanecitrix:
@MarkSymsCtx  have had a look at the bug, hopefully okay now

PR #259, opened 2018-07-13
libvhd: properly check data alignment

cheese:


MarkSymsCtx:
This code is logically dead and unused by the rest of tapdisk. It has been removed in https://github.com/xapi-project/blktap/pull/331.

cheese:
I am sorry to hear that. Removing function from shared library breaks API and ABI. We use this function within our company.

MarkSymsCtx:
Ah, that's a good reason to retain then.

MarkSymsCtx:
Reversion pushed and then we'll take this PR.

MarkSymsCtx (approved):
LGTM

MarkSymsCtx:
@cheese looking at this in more detail, whilst it seems like it works it raises wider questions of alignment throughout libvhd. It maybe that aligning with `pathconf(path,_PC_REC_XFER_ALIGN)` in general is what was originally intended rather than `VHD_SECTOR_SIZE` and this would have an impact on this change of yours as well. As you've obviously been using this function do you have any thoughts around this?

TimSmithCtx (approved):

cheese:
For my understanding, actually buffer alignment is limited by the O_DIRECT flag. Quoted from open(2)

       Under Linux 2.4, transfer sizes, and the alignment of the user buffer and the file offset must all be multiples of the logical block size of  the  filesys‐
       tem.   Since  Linux  2.6.0,  alignment  to  the logical block size of the underlying storage (typically 512 bytes) suffices.  The logical block size can be
       determined using the ioctl(2) BLKSSZGET operation or from the shell using the command:

           blockdev --getss

VHD_SECTOR_SIZE is right the same as the typical logical block size. And I don't know whether `pathconf(path,_PC_REC_XFER_ALIGN)` is the same as the corresponding logical block size.

And the other way, the libvhd code may remove explicit checks of alignment and let the syscalls check.

PR #261, opened 2018-07-30
CA-227096 Allow stashed passed fds to overwrite with the same name

TimSmithCtx:
When a filedescriptor is passed with a name, the name and fd are stored
in a free slot and later looked up by name. This means that if ever a
new fd is passed for a name, it will be invisible as the retriever will
return fd stored against the first matching name, which will be the
first instance.

To fix this, allow passed fds to overwrite the fd stored against an
existing name with the new fd.

Also, the current time was stored against each entry, but this is never
used so remove it. And correct the spelling of
tdnbd_retrieve_passed_fd()

Signed-off-by: Tim Smith <tim.smith@citrix.com>

### DIFF
--- a/drivers/block-nbd.c
+++ b/drivers/block-nbd.c
@@ -74,7 +74,6 @@ struct td_fdreceiver *fdreceiver = NULL;
 
 struct tdnbd_passed_fd {
 	char                    id[40];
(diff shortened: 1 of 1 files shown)


jonludlam:
:+1:

PR #265, opened 2018-09-19
[RFC]: add discard support for vhd

AlexLanzano:
Discard is enabled when the frontend sets feature-discard to 1, the discard-granularity to the device sector size, and the discard-alignment to 0. These values will be written to the xenstore.

Discard is enabled if the following conditions are true:

1.    The device backend supports discard
2.    the device is writable
3.    the device is not a cdrom

How tapdisk handles discard requests:

1.   Requests are pulled off the ring from xenio_blkif_get_request(). Depending on the protocol, 
      the discard request will be converted into a blkif_request_discard.
3.   The request is then converted to a td_vbd_request via tapdisk_xenblkif_make_vbd_request()   
      and the vbd request is queued using tapdisk_vbd_queue_request()
3.   tapdisk_vbd_recheck_state() sees the discard request waiting to be issued. 
4.   tapdisk_vbd_issue_request_discard() packages the vbd request into a td_request 
5.   The request will then be handed off to the block driver to queue the request.

The portion of my patch explained above is a slightly modified version of a previous PR https://github.com/xapi-project/blktap/pull/167/commits/1a51656213d5c7184dbcbfc690e9c7ee026ee96d. Credit goes to Rober Breker for implementing a lot of the discard request handling for tapdisk.

How the vhd block driver handles discard requests:

1.   vhd_queue_discard() receives the request.
2.   The request will be separated into smaller requests where the sectors to be discard would
       be equal to the block length.
3.   The smaller requests will be queued in preallocated request array.
4.   tapdisk_rwio_discard() pulls requests off of the request array and fallocate() is called to 
      deallocate the block.

I had to change the tapdisk-queue driver from using LIO to RWIO because currently linux AIO does not support discard requests. This change causes a 50% sequential write degradation. I've attempted to remedy this by using LIO when there are no discard requests on the queue and revert back to RWIO when a discard is queued. Sadly, the IO scheduler doesn't like when you mix LIO and RWIO.

Are there any other approaches I could take? Is discard request support something you guys are interested in having?

### DIFF
--- a/control/tap-ctl-info.c
+++ b/control/tap-ctl-info.c
@@ -36,7 +36,8 @@
 #include "tap-ctl.h"
 
 int tap_ctl_info(pid_t pid, unsigned long long *sectors,
-		unsigned int *sector_size, unsigned int *info, const int minor)
+				 unsigned int *sector_size, unsigned int *info,
+				 bool *discard, const int minor)
 {
     tapdisk_message_t message;
     int err;
@@ -60,6 +61,7 @@ int tap_ctl_info(pid_t pid, unsigned long long *sectors,
         *sectors = message.u.image.sectors;
         *sector_size = message.u.image.sector_size;
         *info = message.u.image.info;
+		*discard = message.u.image.discard;
         return 0;
     } else if (TAPDISK_MESSAGE_ERROR == message.type) {
        return -message.u.response.error;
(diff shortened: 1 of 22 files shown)


MarkSymsCtx:
Yes, this is interesting but the serious performance degradation would need to be resolved. Given that the current performance of tapdisk isn't as good as we would like degrading it further even with the benefit of trim would be a step in the wrong direction.

PR #267, opened 2018-10-11
Address coverity issues in tapdisk-control.c

MarkSymsCtx:


drivers/tapdisk-control.c
```diff
@@ -1337,7 +1339,7 @@ tapdisk_control_receive_request(struct tapdisk_ctl_conn *conn)
 error:
 	memset(&conn->response, 0, sizeof(conn->response));
 	conn->response.type = TAPDISK_MESSAGE_ERROR;
-	conn->response.u.response.error = (err ? -err : EINVAL);
+	conn->response.u.response.error = -err;
```
MarkSymsCtx: This one requires context of how we can get to the error: label, only by having err set to -EINVAL or -EBUSY

PR #291, opened 2019-09-06
CA-326370 Check return code from io_getevents()

TimSmithCtx:
This can return negative numbers on error, and that can result in very
unexpected behaviour.

Signed-off-by: Tim Smith <tim.smith@citrix.com>

### DIFF
--- a/drivers/tapdisk-queue.c
+++ b/drivers/tapdisk-queue.c
@@ -458,8 +458,14 @@ tapdisk_lio_event(event_id_t id, char mode, void *private)
 	tapdisk_lio_ack_event(queue);
 
 	lio   = queue->tio_data;
-	ret   = io_getevents(lio->aio_ctx, 0,
-			     queue->size, lio->aio_events, NULL);
+	/* io_getevents() invoked via the libaio wrapper does not set errno but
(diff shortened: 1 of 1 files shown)


drivers/tapdisk-queue.c
```diff
@@ -458,8 +458,14 @@ tapdisk_lio_event(event_id_t id, char mode, void *private)
 	tapdisk_lio_ack_event(queue);
 
 	lio   = queue->tio_data;
-	ret   = io_getevents(lio->aio_ctx, 0,
-			     queue->size, lio->aio_events, NULL);
+	/* io_getevents() invoked via the libaio wrapper does not set errno but
+     * instead returns -errno on error */
```
BenSimsCitrix: tab :)

TimSmithCtx: Grrr.

MarkSymsCtx:
> Very educational. I wonder if there are any other system calls via libaio which we are failing to retry on EINTR

Quite possibly because this isn't something the "unchecked return from library" check in coverity will catch as we did actually assign the return but then not do correct validation on it.

PR #293, opened 2019-10-10
Blktap performance updates

MarkSymsCtx:
* Enable multipage ring support in blktap
* When in polling mode, mask the event channel events to reduce system call overhead.

### DIFF
--- a/include/xen_blkif.h
+++ b/include/xen_blkif.h
@@ -130,4 +130,9 @@ static inline void blkif_get_x86_64_req(blkif_request_t *dst, blkif_x86_64_reque
 		dst->seg[i] = src->seg[i];
 }
 
+#define MAX_RING_PAGE_ORDER 3
+#define MAX_RING_PAGES (1 << MAX_RING_PAGE_ORDER)
+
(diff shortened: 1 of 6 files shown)


include/xen_blkif.h
```diff
@@ -130,4 +130,9 @@ static inline void blkif_get_x86_64_req(blkif_request_t *dst, blkif_x86_64_reque
 		dst->seg[i] = src->seg[i];
 }
 
+#define MAX_RING_PAGE_ORDER 3
```
BenSimsCitrix: There is a reference to BLKIF_MAX_SEGMENTS_PER_REQUEST in drivers/td-ctx.c used in the blkif_get_req macro, does this need to change too?

BenSimsCitrix: Damn its on the next page! Ignore me!

tapback/backend.c
```diff
@@ -39,6 +39,7 @@
 #include <unistd.h>
```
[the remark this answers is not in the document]
MarkSymsCtx: We could but I'd expect it to have the correct include once syntax and so this is probably safer rather than relying on a second level include.

PR #294, opened 2019-10-10
LIBBLOCKCRYPTO_NAME consolidation & versioning

jandryuk:
This PR consolidates the three definitions of LIBBLOCKCRYPTO_NAME into libvhd.h and adds the major version (.0) to the string.

OpenXT doesn't install a `libblockcrypto.so` symlink, only `libblockcrypto.so.0` as a symlink to `libblockcrypto.so.0.0.0`.  This leads to a runtime failure when dlopen can't find the library - "failed to load crypto support library".  It's more defensive to specify the major version.  That way a major version change will result in an error instead of being used when there can be incompatible changes.

`libblockcrypto.so` should already be a symlink to `libblockcrypto.so.0.0.0` anyway, so there should be no functional change.  I've tested this on OpenXT, but not XenServer.  On an existing XenServer system, you can verify the `libblockcrypto.so.0` is present to avoid this PR breaking blktap.

### DIFF
--- a/drivers/block-vhd.c
+++ b/drivers/block-vhd.c
@@ -79,8 +79,6 @@
 
 unsigned int SPB;
 
-#define LIBBLOCKCRYPTO_NAME "libblockcrypto.so"
-
 #define DEBUGGING   2
 #define MICROSOFT_COMPAT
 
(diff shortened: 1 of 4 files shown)


TimSmithCtx:
> Thanks. @TimSmithCtx this doesn't change the rpm packaging. I'm not really in a place to try that out, but it may be as easy as moving the `%{_libdir}/*.so` line

Yes, it's easy to fix. We just need to remember to do it :-)

PR #296, opened 2019-10-24
CA-327558: use preadv/pwritev to merge IO

edwintorok:
The current IO merging requires both contiguous offsets and contiguous buffers in memory.
The latter occurs very rarely due to the way buffers are put back on IO completion (most often it ends up in reverse order, or with large gaps).
`preadv/pwritev` only requires contiguous offsets, so there is more opportunity for merging.

Together with https://github.com/xapi-project/blktap/pull/295 it yield ~27% improvement.

To be able to add pread/pwritev support I first had to drop a lot of dead code (so I won't have to update it), and to avoid directly accessing the union members of `iocb`. These have to be accessed using helper functions because `preadv/pwritev` use a different union member.


drivers/io-optimize.h
```diff
@@ -42,7 +42,10 @@ struct opio_list {
 	struct opio        *tail;
 };
 
+#define UIO_FASTIOV 8
```
MarkSymsCtx: Is this a hardcoded value in the kernel or could it be configured? i.e. should we/could we be reading this from the kernel?

edwintorok: It is present in `linux/uio.h`, but if we include that we get a double definition of `struct iovec` between `linux/uio.h` and `bits/uio.h` included through `fcntl.h`

PR #297, opened 2019-10-28
CA-327558: avoid some eventfd reads

edwintorok:
When we're polling in user-space this is not needed.
Gives ~4% improvement on bs 2M, iodepth 8.

Also delete obsolete code when eventfd is not available (in our kernels it always is).

Signed-off-by: Edwin Török <edvin.torok@citrix.com>

### DIFF
--- a/drivers/tapdisk-queue.c
+++ b/drivers/tapdisk-queue.c
@@ -188,11 +188,13 @@ struct lio {
 
 	int              event_fd;
 	int              event_id;
+	int              event_poll_id;
 
 	int              flags;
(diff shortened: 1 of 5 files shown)


edwintorok:
I tried to convert `select()` into `IOCB_CMD_POLL` (which would've allowed us to poll entirely from user-space), but I got EINVAL telling me that polling is not supported on some of the fds (or I got something else wrong with that), so for now we'll keep the select.
Perhaps `io_uring` would give us a better solution when it is available.

PR #301, opened 2020-02-05
CP-32123: Integrate blktap with CCM openssl 1.1.1

minglumlu:
This patch needs to be merged with blktap.spec together as CCM openssl
1.1.1 is required when building this package

Signed-off-by: Qin Zhang <qin.zhang@citrix.com>


MarkSymsCtx:
Can't be merged until the build system is ready with the build requirements.

MarkSymsCtx:
Please add a 
```
BuildRequires: openssl-devel >= 1.1
```
to this https://github.com/xapi-project/blktap/blob/master/mk/blktap.spec.in#L13 to make it clear to any external consumer that it requires OpenSSL 1.1 or later

PR #313, opened 2020-04-28
CA-338603: fix out of band write in tapdisk-control

MarkSymsCtx:


MarkSymsCtx:
> There is another instance of this on line 742, maybe zap that one as well while we are at it, or pick it up in separate ticket.

Well spotted, that isn't guaranteed to terminate either so fixed that too.

PR #318, opened 2020-06-09
feat(vhd/canonpath): add support for DRBD resource path

Wescoeur:
Preserve symlinks when "/dev/drbd/by-res/<RES_NAME>/<VOLUME>" paths are used.
It's interesting when a parent is set to a VHD file to avoid usage of
"/dev/drbdXXXX" paths where XXXX is the device minor number.

We can have:
>\> vhd-util query -n /dev/drbd/by-res/volume-d803817e-6113-4b89-abdf-d4ab39d63e8c/0 -p
/dev/drbd/by-res/volume-d5f5a079-0529-4b97-95be-9a859056d523/0

instead of:
>\> vhd-util query -n /dev/drbd/by-res/volume-d803817e-6113-4b89-abdf-d4ab39d63e8c/0 -p
/dev/drbd/./drbd1004

If a DRBD resource exists on many hosts with a resource name equal to an UUID, it can be used
independently of the DRBD device minor number on each host.

Signed-off-by: Ronan Abhamon <ronan.abhamon@vates.fr>

### DIFF
--- a/vhd/lib/libvhd.c
+++ b/vhd/lib/libvhd.c
@@ -121,7 +121,8 @@ old_clear_bit(volatile char *addr, int nr)
 }
 
 static int
-makedev_from_file(const char *file, dev_t *dev) {
+makedev_from_file(const char *file, dev_t *dev)
+{
 	*dev = 0;
 
 	FILE *f = fopen(file, "r");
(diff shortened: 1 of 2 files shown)


Wescoeur:
@TimSmithCtx Hi, could you take a look at this commit please?
It's a particular feature but useful to read correctly DRBD devices on multiple hosts without following symlinks. (There is already a specific code concerning the LVM devices in the canonpath module, I'm just adding another specific piece of code here.)

Wescoeur:
@TimSmithCtx @MarkSymsCtx Hi! Do you have some time to read this PR? I tested this patch, it's a little intrusive but I would like to see it merged to support DRBD disks correctly.

vhd/lib/libvhd.c
```diff
@@ -119,6 +120,128 @@ old_clear_bit(volatile char *addr, int nr)
...
+		#ifdef _DIRENT_HAVE_D_TYPE
+			if (ent->d_type == DT_LNK) {
+				/* Ok here, we have a symlink! */
+			} else if (ent->d_type != DT_UNKNOWN)
+				continue;
+			else
+		#endif
+		{
+			 struct stat slave_stats;
+			 stat(ent->d_name, &slave_stats);
```
MarkSymsCtx, line 217: You don't check the return value of stat here, this fails a Coverity check as the stat struct may not have been populated.

"check_return: Calling stat(ent->d_name, &slave_stats) without checking return value. This library function may fail and return an error code"

Wescoeur: Oops, yes that's right. I just fixed it!

Wescoeur:
@MarkSymsCtx Hi! Do you have any concerning this PR?

MarkSymsCtx:
No, just busy with higher priority things at the moment.

MarkSymsCtx (changes requested):
Fix normalize_path in line with comments from @TimSmithCtx

Wescoeur:
@MarkSymsCtx @TimSmithCtx I fixed the `normalize_path` function, it returns an allocated string now. Valgrind seems happy with invalid, relative and absolute paths.

TimSmithCtx (approved):

vhd/lib/libvhd.c
```diff
@@ -119,6 +120,129 @@ old_clear_bit(volatile char *addr, int nr)
...
+		return -1;
+
+	*dev = makedev(dev_major, dev_minor);
+	return 0;
+}
+
+static int
+is_drbd_device(const struct stat *stats)
+{
+	static const int drbd_major = 147;
```
MarkSymsCtx, line 147: Is this guaranteed to always be 147? There is other code that looks up device major ids from `/proc/devices`

Wescoeur: Normally yes, like said in the official DRBD documentation: https://www.linbit.com/drbd-user-guide/users-guide-drbd-8-4/

> DRBD device
>
> This is a virtual block device managed by DRBD. It has a device major number of 147, and its minor numbers are numbered from 0 onwards, as is customary. Each DRBD device corresponds to a volume in a resource.

>  Very early DRBD versions hijacked NBD’s device major number 43. This is long obsolete; 147 is the LANANA-registered DRBD device major.

MarkSymsCtx: Yes, thanks, I found the assigned number. I'm just more used to the code in this area dealing with the blktap devices which use the experimental range (and it looks like they can move around a bit).

PR #319, opened 2020-06-22
Private/bensi/cp 32853

BenSimsCitrix:


drivers/posixaio-backend.c
```diff
@@ -0,0 +1,598 @@
...
+ */
+static void
+complete_tiocb(posix_aio_queue *queue, struct tiocb *tiocb)
+{
+	int err;
+	unsigned long actual_res;
+	struct aiocb *aiocb = tiocb->iocb;
+	unsigned long res = aiocb->aio_nbytes;
+
+	/*TO DO THIS IS WRONG IN NORMAL QUEUE*/
```
MarkSymsCtx, line 207: Does this need a fix somewhere?

drivers/posixaio-backend.c
```diff
@@ -0,0 +1,598 @@
...
+#include <unistd.h>
+#include <aio.h>
+
+#include "tapdisk.h"
+#include "tapdisk-log.h"
+#include "posixaio-backend.h"
+#include "tapdisk-server.h"
+#include "tapdisk-utils.h"
+#include "timeout-math.h"
+#include <signal.h>
```
edwintorok, line 47: Nit: might be nice to have all system includes grouped together, and then all local includes.

drivers/posixaio-backend.c
```diff
@@ -0,0 +1,598 @@
...
+#include "aio_getevents.h"
+#include <sys/signalfd.h>
+#include "debug.h"
+
+
+#define WARN(_f, _a...) tlog_write(TLOG_WARN, _f, ##_a)
+#define DBG(_f, _a...) tlog_write(TLOG_DBG, _f, ##_a)
+#define ERR(_err, _f, _a...) tlog_error(_err, _f, ##_a)
+
+#define posixaio_backend_queue_empty(q) ((q)->queued == 0)
```
edwintorok, line 59: Nit: these could be `static inline`, and then you'd get some type checking

drivers/posixaio-backend.c
```diff
@@ -0,0 +1,598 @@
...
+};
+
+enum {
+	TIO_DRV_LIO     = 1,
+};
+
+static inline void
+queue_tiocb(posix_aio_queue *queue, struct tiocb *tiocb)
+{
+	queue->aiocbList[queue->queued] = (struct aiocb*)tiocb->iocb;
```
edwintorok, line 97: Might be useful to have a bounds check here, at least as an assert. One of the callers does a check whether the queue is full or not before calling or not, but the other one doesn't seem to.

drivers/posixaio-backend.c
```diff
@@ -0,0 +1,598 @@
...
+}
+
+static inline int
+deferred_tiocbs(posix_aio_queue *queue)
+{
+	return (queue->deferred.head != NULL);
+}
+
+static inline void
+defer_tiocb(posix_aio_queue *queue, struct tiocb *tiocb)
```
edwintorok, line 109: I'm not sure why deferring is needed. Posix AIO should be able to handle any number of items queued (and there is a limit anyway when the rings become full and there is no more room on where to put the replies).
Could the code here be simplified by storing whatever context we need as part of the IO request itself? If the goal was to avoid OS memory allocation during the fastpath that could be solved with a generic memory pool from which we can take and return memory, without complicating the main logic of the code.

edwintorok (approved):
I don't think that any of my oustanding review comments should block the merge.
But would be good if they could be addressed either here or in a followup PR @BenSimsCitrix

PR #320, opened 2020-07-03
Coverity Fixes

MarkSymsCtx:


drivers/scheduler.c
```diff
@@ -115,17 +115,17 @@ scheduler_prepare_events(scheduler_t *s)
 		if (event->masked || event->dead)
 			continue;
 
-		if (event->mode & SCHEDULER_POLL_READ_FD) {
+		if (event->mode & SCHEDULER_POLL_READ_FD && event->fd > 0) {
```
TimSmithCtx, line 118: Should this be "eventfd >= 0" ?

(likewise below)

vhd/lib/vhd-util-resize.c
```diff
@@ -207,6 +207,7 @@ swap(vhd_block_t *list, int a, int b)
 	list[b] = tmp;
 }
 
+#if 0
 static int
```
BenSimsCitrix, line 211: Why no just delete it?

drivers/scheduler.c
```diff
@@ -115,17 +115,17 @@ scheduler_prepare_events(scheduler_t *s)
...
 			FD_SET(event->fd, &s->read_fds);
 			s->max_fd = MAX(event->fd, s->max_fd);
 		}
 
-		if (event->mode & SCHEDULER_POLL_WRITE_FD) {
```
BenSimsCitrix, line 123: my 2p 
I prefer if ((blah1 & blah2) && blah3), makes intention clearer even though operator precedence ensures correct result.

drivers/block-nbd.c
```diff
@@ -680,6 +680,7 @@ tdnbd_connect_import_session(struct tdnbd_data *prv, td_driver_t* driver)
 			sizeof(opt));
 	if (rc < 0) {
 		ERROR("Could not set TCP_NODELAY: %s\n", strerror(errno));
+		close(sock);
 		return -1;
 	}
```
BenSimsCitrix, line 685: Not your code, but do we need to free the prv->remote on line 716

drivers/scheduler.c
MarkSymsCtx: yes

vhd/lib/vhd-util-resize.c
MarkSymsCtx: Because there's a lot of code here and we might want to revisit whether it can be made to work safely. Deleting it make it more difficult to see that there is some history here and maybe a starting point.

Normally I'd agree on deleting unused code but I think this justifies being a special case..

drivers/block-nbd.c
MarkSymsCtx: Looks like it. Will check why we don't get a resource leak event there.

drivers/scheduler.c
MarkSymsCtx: Done

drivers/block-nbd.c
MarkSymsCtx: Done

drivers/scheduler.c
MarkSymsCtx: Done

PR #324, opened 2020-07-31
CP-32988 - Remove cgroup manipulation from tap-ctl.

BenSimsCitrix:
Revert "CA-225067: tap_ctl_{spawn,create}: move tapdisk to cgroup slice"
This reverts commit ce3df05d75310f130928e391fe17e7755809dd42.

Signed-off-by: ben sims <ben.sims@citrix.com>


MarkSymsCtx:
I think we should have a more detailed commit comment for this so that posterity will understand why we're moving this code into the calling control plane.

PR #326, opened 2020-08-12
CA-343027: reorder event channel unmasking to avoid possible race

MarkSymsCtx:


MarkSymsCtx:
this doesn't seem to have been the actual issue with the storage stall as the masking/unmasking just controls the fdset supplied to select.

PR #333, opened 2020-09-07
CA-344119: return correct error, snprintf doesn't set errno

MarkSymsCtx:

### DIFF
--- a/vhd/lib/libvhd.c
+++ b/vhd/lib/libvhd.c
@@ -2717,7 +2717,8 @@ vhd_open(vhd_context_t *ctx, const char *file, int flags)
 		struct stat stats;
 		err = stat(file, &stats);
 		if (err == -1)
-			return -errno;
+			/* Coverity seems not understand that stat sets errno. */
+			return errno ? -errno : -EINVAL;
 
 		/* If a DRBD path is used, we try to use the real physical device. */
 		/* Why? Because the device can be locked by any program. */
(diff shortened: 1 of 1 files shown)


MarkSymsCtx:
@Wescoeur This issue was introduced in your recent changes. Please take look.

Wescoeur:
@MarkSymsCtx Ok! Your patch looks good to me.

vhd/lib/libvhd.c
```diff
@@ -2717,7 +2717,7 @@ vhd_open(vhd_context_t *ctx, const char *file, int flags)
 		struct stat stats;
 		err = stat(file, &stats);
 		if (err == -1)
-			return -errno;
+			return errno ? -errno : -EINVAL;
```
TimSmithCtx, line 2720: Worth a comment here that this is necessary because Coverity doesn't understand that stat() sets errno? Otherwise someone might later read the man page for stat() and "clean it up"

MarkSymsCtx: Of course Synopsis might fix that as well and then it can be cleaned up.

vhd/lib/libvhd.c
```diff
@@ -2726,7 +2727,7 @@ vhd_open(vhd_context_t *ctx, const char *file, int flags)
 		dev_t mapper_dev;
 		if (drbd_is_up_to_date(&stats) && drbd_to_mapper(&stats, &mapper_dev) == 0) {
 			if (snprintf(bypass_file, sizeof bypass_file, "/dev/block/%u:%u", major(mapper_dev), minor(mapper_dev)) == -1)
```
rosslagerwall, line 2729: Not really related to this PR, but should it also check for `snprintf() >= sizeof(bypass_file)`?

rosslagerwall (approved):

vhd/lib/libvhd.c
MarkSymsCtx: Quite probably but I'll let XenOrchestra worry about that.

Wescoeur: `bypass_file` is big enough to contain `snprintf(...)` characters in this case :wink:: https://github.com/xapi-project/blktap/blob/74ecccfb33a59bc1ca8ade9a8f85070ed7bda0e8/vhd/lib/libvhd.c#L2715

BenSimsCitrix (approved):
LGTM

PR #334, opened 2020-09-10
CA-344424: ensure we don't get division by 0

MarkSymsCtx:
Another coverity find.

### DIFF
--- a/cpumond/cpumond.c
+++ b/cpumond/cpumond.c
@@ -38,6 +38,7 @@
 #include <fcntl.h>
 #include <unistd.h>
 #include <errno.h>
+#include "compiler.h"
 #include "cpumond.h"
 
 #ifndef DEBUG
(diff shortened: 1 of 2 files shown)


cpumond/cpumond.c
```diff
@@ -170,6 +171,9 @@ int cpumond_loop(cpumond_entry_t *cpumond_entry){
         if (statread(statfd, &total2, &idle2) != 0)
             goto out;
 
+	if (unlikely(total2 == total1))
```
MarkSymsCtx, line 174: Actually, this should go to the sleep call

TimSmithCtx (approved):

cpumond/cpumond.c
```diff
@@ -170,6 +171,11 @@ int cpumond_loop(cpumond_entry_t *cpumond_entry){
         if (statread(statfd, &total2, &idle2) != 0)
             goto out;
 
+	if (unlikely(total2 == total1)) {
+		sleep(1);
```
BenSimsCitrix, line 175: Don't you need to reread total2?

BenSimsCitrix: Ahh i see now this is within a loop.

BenSimsCitrix (approved):
LGTM

cpumond/cpumond.c
MarkSymsCtx: Yes, and I used continue with a second sleep as wrapping the update code in a ```if (total2 > total2) {``` still resulted iin Coverity moaning about untrusted divisors :( stupid tool at times.

PR #335, opened 2020-09-22
CA-345193: allocate sufficient space in normalize_path

MarkSymsCtx:
Signed-off-by: Mark Syms <mark.syms@citrix.com>

PR #336, opened 2020-09-22
Add a unit test for canonpath that exercises the code affected by CA-345193

MarkSymsCtx:
If the fix in https://github.com/xapi-project/blktap/pull/335 is reverted in the presence of this unit test it will correctly fail with a buffer overrun error.


TimSmithCtx (approved):

BenSimsCitrix (approved):
LGTM

PR #344, opened 2021-03-09
CA-35250 - Fix Coverity issues introduced by CP-32853

BenSimsCitrix:
Running this through a BST. I will remove the do not merge if reviewers are happy and the BST succeeds.

I used a union to avoid the malloc and void* (Coverity complained that the malloc could fail and we'd have a NULL pointer) Propagating a failure in malloc would have meant changing function interfaces in many files and very error prone.

### DIFF
--- a/drivers/libaio-backend.c
+++ b/drivers/libaio-backend.c
@@ -110,7 +110,7 @@ struct tio {
 static inline void
 queue_tiocb(libaio_queue *queue, struct tiocb *tiocb)
 {
-	struct iocb *iocb = (struct iocb*)tiocb->iocb;
+	struct iocb *iocb = &(tiocb->uiocb.io);
 
 	if (queue->queued) {
 		struct tiocb *prev = (struct tiocb *)
@@ -172,7 +172,7 @@ static void
 complete_tiocb(libaio_queue *queue, struct tiocb *tiocb, unsigned long res)
 {
 	int err;
-	struct iocb *iocb = (struct iocb*)tiocb->iocb;
+	struct iocb *iocb = &(tiocb->uiocb.io);
 
 	if (res == iocb_nbytes(iocb))
 		err = 0;
@@ -182,7 +182,6 @@ complete_tiocb(libaio_queue *queue, struct tiocb *tiocb, unsigned long res)
 		err = -EIO;
 
 	tiocb->cb(tiocb->arg, tiocb, err);
-	free(iocb);
 }
 
 static int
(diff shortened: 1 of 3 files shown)


drivers/libaio-backend.c
```diff
@@ -110,7 +110,7 @@ struct tio {
 static inline void
 queue_tiocb(libaio_queue *queue, struct tiocb *tiocb)
 {
-	struct iocb *iocb = (struct iocb*)tiocb->iocb;
+	struct iocb *iocb = (struct iocb*)&(tiocb->uiocb);
```
MarkSymsCtx, line 113: This should be `(struct iocb*)&(tiocb->uiocb.io)`, similarly where the aiocb is referenced `(struct aiocb*)&(tiocb->uiocb.aiocb)`. In pointer terms it almost certainly equates to the same thing, both members of the union being aling with the start of the union but it's not correct.

MarkSymsCtx (changes requested):

drivers/libaio-backend.c
edwintorok: You can remove the cast actually in this case, since the union member should have the correct type already (prevents accidentally casting it to a wrong type).

drivers/posixaio-backend.c
```diff
@@ -486,13 +485,13 @@ posixaio_backend_init_queue(tqueue *pqueue, int size,
 		goto fail;
 	}
 
-	queue->aiocbList = calloc(size, sizeof(struct aiocb));
+	queue->aiocbList = calloc(size, sizeof(struct aiocb*));
```
edwintorok, line 488: Would `calloc(size, sizeof(queue->aiocbList[0]))` be more reliable? it'd guarantee the types match.

drivers/libaio-backend.c
BenSimsCitrix: Seen many unions in code but actually used one anger, so i've learned a lot here. I will fix this up and retest

drivers/posixaio-backend.c
MarkSymsCtx: I don't think that's correct. This calloc is allocating an array of pointers to `struct aiocb` not an array of `struct aiocb`

drivers/libaio-backend.c
BenSimsCitrix: Fix in latest commit

drivers/posixaio-backend.c
BenSimsCitrix: I will leave this one in place then, it seems to be the style generally in blktap.

Interestingly Edvins suggestion does compile though, I wonder if the compiler is really opening an uninitalized array or if it is doing some smart type inference during compile time.

MarkSymsCtx: Actually, it's probably the same thing if queue->aiocblist is a `struct aiocb **`

BenSimsCitrix:
Rerunning bst suite run 145458

TimSmithCtx (approved):

BenSimsCitrix:
Coverity is happy on my local and storage BST is green.

PR #350, opened 2021-06-29
Improve error reporting from vhd-util scan

MarkSymsCtx:


vhd/lib/vhd-util-scan.c
```diff
@@ -1255,13 +1255,17 @@ vhd_util_scan_find_volume_targets(int cnt, char **names,
 	targets   = NULL;
 
 	err = lvm_scan_vg(volume, &vg);
-	if (err)
+	if (err) {
+		fprintf(stderr, "lvm_scan_vg failed %s\n", err);
```
edwintorok, line 1259: shouldn't this be %d?

edwintorok (changes requested):

vhd/lib/vhd-util-scan.c
MarkSymsCtx: Yes and it failed to compile, I obviously updated this after the last compile. Fixed now.

TimSmithCtx (approved):

edwintorok (approved):

vhd/lib/vhd-util-scan.c
edwintorok: might be useful to add some github actions to try and build blktap, but getting all the dependencies would be tricky, they probably won't be neatly available in distro packages.

MarkSymsCtx: The dependencies won't be available at all in the distros, as they require the frankentap kernel device and headers which only exist in the Citirx Hypervisor kernel patchqueue.

PR #351, opened 2021-07-14
CA-356508: handle EINTR in tap-ctl IPC

MarkSymsCtx:


edwintorok (commented):
There is a `select` a few lines above, that could use similar EINTR treatment.

edwintorok (approved):
LGTM

rosslagerwall (approved):

TimSmithCtx (approved):

PR #356, opened 2022-03-08
Fix new style nbdserver support

jandryuk:
These patches get nbd-client to be able to connect to tapdisk's nbdserver in new style mode.  nbd-client dropped old style support, so it cannot connect to the existing nbdserver.  To test this, I patched drivers/tapdisk-nbdserver.c to set `SERVER_USE_OLD_PROTOCOL = 0`.

`nbd-client -unix /run/blktap-control/nbd$PID.$MINOR /dev/nbd0` now connects (with SERVER_USE_OLD_PROTOCOL = 0).

### DIFF
--- a/drivers/tapdisk-nbdserver.c
+++ b/drivers/tapdisk-nbdserver.c
@@ -825,7 +825,7 @@ tapdisk_nbdserver_handshake_cb(event_id_t id, char mode, void *data)
 	}
 }
 
-static void
+void
 tapdisk_nbdserver_newclient_fd_new_fixed(td_nbdserver_t *server, int new_fd)
 {
 	ASSERT(server);
--- a/drivers/tapdisk-nbdserver.c
+++ b/drivers/tapdisk-nbdserver.c
@@ -244,12 +244,12 @@ tapdisk_nbdserver_reqs_free(td_nbdserver_client_t *client)
 }
 
 void
-provide_server_info (td_nbdserver_t *server, uint64_t *exportsize, uint16_t *flags)
+provide_server_info (td_nbdserver_client_t *client, uint64_t *exportsize, uint16_t *flags)
 {
 	int64_t size;
 	uint16_t eflags = NBD_FLAG_HAS_FLAGS;
 
-	size = server->info.size * server->info.sector_size;
+	size = client->server->info.size * client->server->info.sector_size;
 
 	*exportsize = size;
 	*flags = eflags;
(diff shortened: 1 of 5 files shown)


MarkSymsCtx:
@jandryuk does this still permit the old style NBD connection without needing to enforce a switch on the tapdisk side? Ben Sims who made these original sets of changes to support the new style handshake found that he couldn't get it to work on both handshake types and this caused product regressions around things like Storage Migration, particularly when running between product versions where it's not possible to update an down version install. Obviously OpenXT doesn't need to concern itself on that but in the general case we absolutely do.

MarkSymsCtx:
The change does however fail the tests executed by `make check` which are called by the %check step in the spec file.

jandryuk:
Of course.  Sorry about that.

edwintorok:
Will probably need some changes on the XAPI side to use a new URL for storage migration that supports the new protocol, and the old URL would keep forwarding to tapdisk with the old protocol.

It looks like XAPI forwards the migration stream to tapdisk here: https://github.com/xapi-project/xen-api/blob/47fae74032aa6ade0fc12e867c530eaf2a96bf75/ocaml/xapi/storage_migrate.ml#L725-L743 to send a control message to tapdisk asking for new-style NBD,
and a change here to use a new URL for it: https://github.com/xapi-project/xen-api/blob/47fae74032aa6ade0fc12e867c530eaf2a96bf75/ocaml/xapi/storage_migrate.ml#L705-L710
(and coresponding code in xapi to handle the new URL on the receiving side).
Because we only support migrating from older versions to versions >= our version I think we can always *initiate* a migration using new style NBD, but we have to *accept* migrations using both protocols.

MarkSymsCtx:
> cmocka is being difficult. The LOG statements in tapdisk_nbdserver_newclient_fd_new_fixed call syslog which eventually calls send. This makes using __wrap_send tricky (at least to a cmocka novice like me).

@jandryuk we're under a deadline crunch on something at the moment but when I get out from under that I'll try and take a look at that as I did most of the original cmocka stuff and it's a bit fiddly. Might be a case of mocking out tlog_syslog itself, so a `-Wl,--wrap=tlog_syslog` added here - https://github.com/xapi-project/blktap/blob/master/mockatests/drivers/Makefile.am#L15 maybe ? And then a no-op wrapper function in here https://github.com/xapi-project/blktap/blob/master/mockatests/drivers/vbd-wrappers.c#L39

jandryuk:
Thanks, @MarkSymsCtx .  There's no rush on my end, so good luck with your deadline.

jandryuk:
Adding the linker wrap command was important.  I used __wrap_tapdisk_syslog since I started using that before you mentioned tlog_syslog.

It's not pretty and is fiddly since there are syslog calls in the called functions.

The other option is to factor out a new tapdisk_nbdserver_new_protocol_handshake() without any LOG statements.
Then the old test could be basically reused.  But it seems weird to change
the code to make writing the test easier.

jandryuk:
@MarkSymsCtx Looks like your recent NBD work superseded this, so I'm closing this.

PR #357, opened 2022-03-11
GCC 11.2 compile fixes

jandryuk:
This series is assorted compilation fixes for GCC 11.2 (Fedora 35).

### DIFF
--- a/drivers/tapdisk-utils.h
+++ b/drivers/tapdisk-utils.h
@@ -100,6 +100,6 @@ shm_create(struct shm *shm);
 int
 shm_destroy(struct shm *shm);
 
-inline long long timeval_to_us(struct timeval *tv);
+long long timeval_to_us(struct timeval *tv);
 
 #endif
--- a/mockatests/wrappers/wrappers.c
+++ b/mockatests/wrappers/wrappers.c
@@ -96,7 +96,7 @@ int
 __wrap_fwrite(const void *ptr, size_t size, size_t nmemb, FILE *stream)
 {
 	if (mock_fwrite) {
-		struct fwrite_data *data = mock();
+		struct fwrite_data *data = (struct fwrite_data *)mock();
 
 		size_t remaining = data->size - data->offset;
 		size_t len = size * nmemb;
@@ -142,7 +142,7 @@ int
 wrap_vprintf(const char *format, va_list ap)
 {
 	if (mock_vprintf) {
-		struct printf_data *data = mock();
+		struct printf_data *data = (struct printf_data *)mock();
 		int remaining = data->size - data->offset;
 		int len = vsnprintf(data->buf + data->offset, remaining, format, ap);
 		assert_in_range(len, 0, remaining);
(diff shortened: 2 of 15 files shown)


edwintorok (commented):
I think a lot of the printf buffer size warnings won't be reached in practice due to other limits on the numbers used in formatting (that the compiler doesn't know about), but it is better to fix these warnings and be sure that we always fit in the allocated string, even if the numbers happen to be outside the range we're expecting.
vbd->name appears to be a filename, so this should be entirely Dom0 controlled.

edwintorok (commented):
(comment moved into appropriate commit)

mockatests/wrappers/wrappers.c
```diff
@@ -96,7 +96,7 @@ int
 __wrap_fwrite(const void *ptr, size_t size, size_t nmemb, FILE *stream)
 {
 	if (mock_fwrite) {
-		struct fwrite_data *data = mock();
+		struct fwrite_data *data = (struct fwrite_data *)mock();
```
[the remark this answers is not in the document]
jandryuk: Thanks, @edwintorok .  This seems to work:
`struct printf_data *data = mock_ptr_type(struct printf_data *);`

MarkSymsCtx:
Apologies @jandryuk just getting chance to look at this. Having updated it locally on top of master it builds when using `devtookit-11`  so long as `-Wno-stringop-truncation` is specified as it seems to not correctly understand whether strings are being truncated and/or whether this is intentional or not. Based on this I think it's ok to merge, pending the results of some internal tests which I have running and when the others have reviewed it

MarkSymsCtx (approved):

mockatests/wrappers/wrappers.c
```diff
@@ -142,7 +142,7 @@ int
 wrap_vprintf(const char *format, va_list ap)
 {
 	if (mock_vprintf) {
-		struct printf_data *data = mock();
+		struct printf_data *data = (struct printf_data *)mock();
```
rosslagerwall, line 145: Should this also use `mock_ptr_type()`?

drivers/tapdisk-utils.h
```diff
@@ -100,6 +100,6 @@ shm_create(struct shm *shm);
 int
 shm_destroy(struct shm *shm);
 
-inline long long timeval_to_us(struct timeval *tv);
+long long timeval_to_us(struct timeval *tv);
```
andyhhp, line 103: Ewwww.  The way to do this nicely is to delete everything from tapdisk-util.c and have

```c
static inline uint64_t timeval_to_us(const struct timeval *tv)
{
    return (uint64_t)tv->tv_sec * 1000000 + tv->tv_usec;
}
```

Calling into a separate TU for a multiplication like this is mad.

rosslagerwall (approved):

mockatests/wrappers/wrappers.c
MarkSymsCtx: Yes, probably but I'll catch that in some follow-ups I think

TimSmithCtx (approved):

PR #359, opened 2022-05-05
Fix a block of newly detected Coverity issues

MarkSymsCtx:
Coverity 2021.12.1 adds complaints about strcpy and sprintf, so fix them.

### DIFF
--- a/control/tap-ctl-unpause.c
+++ b/control/tap-ctl-unpause.c
@@ -59,8 +59,8 @@ tap_ctl_unpause(const int id, const int minor, const char *params, int flags,
 			     sizeof(message.u.params.path));
 	if (secondary) {
 		err = snprintf(message.u.params.secondary,
-				sizeof(message.u.params.secondary) - 1, "%s",
-				secondary);
+			       sizeof(message.u.params.secondary), "%s",
+			       secondary);
 		if (err >= sizeof(message.u.params.secondary)) {
 			EPRINTF("secondary image name too long\n");
 			return -ENAMETOOLONG;
--- a/control/tap-ctl-unpause.c
+++ b/control/tap-ctl-unpause.c
@@ -40,6 +40,7 @@
 #include <getopt.h>
 
 #include "tap-ctl.h"
+#include "util.h"
 
 int
 tap_ctl_unpause(const int id, const int minor, const char *params, int flags,
@@ -54,8 +55,8 @@ tap_ctl_unpause(const int id, const int minor, const char *params, int flags,
 	message.u.params.flags = flags;
 
 	if (params)
-		strncpy(message.u.params.path, params,
-				sizeof(message.u.params.path) - 1);
+		safe_strncpy(message.u.params.path, params,
+			     sizeof(message.u.params.path));
 	if (secondary) {
 		err = snprintf(message.u.params.secondary,
 				sizeof(message.u.params.secondary) - 1, "%s",
--- a/drivers/block-nbd.c
+++ b/drivers/block-nbd.c
@@ -51,6 +51,7 @@
 #include "timeout-math.h"
 #include "tapdisk-nbdserver.h"
 #include "tapdisk-protocol-new.h"
+#include "util.h"
 
 #ifdef HAVE_CONFIG_H
 #include "config.h"
@@ -199,9 +200,8 @@ tdnbd_stash_passed_fd(int fd, char *msg, void *data)
 		close(passed_fds[free_index].fd);
 
 	passed_fds[free_index].fd = fd;
-	strncpy(passed_fds[free_index].id, msg,
-		sizeof(passed_fds[free_index].id) - 1);
-	passed_fds[free_index].id[sizeof(passed_fds[free_index].id) - 1] = '\0';
+	safe_strncpy(passed_fds[free_index].id, msg,
+		     sizeof(passed_fds[free_index].id));
 }
 
 static int
@@ -921,8 +921,7 @@ tdnbd_open(td_driver_t* driver, const char* name,
 			return -1;
 		}
 		prv->remote_un.sun_family = AF_UNIX;
-		strncpy(prv->remote_un.sun_path, name, sizeof(prv->remote_un.sun_path));
-		prv->remote_un.sun_path[sizeof(prv->remote_un.sun_path) - 1] = '\0';
+		safe_strncpy(prv->remote_un.sun_path, name, sizeof(prv->remote_un.sun_path));
 		len = strlen(prv->remote_un.sun_path)
 			+ sizeof(prv->remote_un.sun_family);
 		if ((rc = connect(prv->socket, (struct sockaddr*)&prv->remote_un, len)
--- a/drivers/block-nbd.c
+++ b/drivers/block-nbd.c
@@ -921,7 +921,8 @@ tdnbd_open(td_driver_t* driver, const char* name,
 			return -1;
 		}
 		prv->remote_un.sun_family = AF_UNIX;
-		strcpy(prv->remote_un.sun_path, name);
+		strncpy(prv->remote_un.sun_path, name, sizeof(prv->remote_un.sun_path));
+		prv->remote_un.sun_path[sizeof(prv->remote_un.sun_path) - 1] = '\0';
 		len = strlen(prv->remote_un.sun_path)
 			+ sizeof(prv->remote_un.sun_family);
 		if ((rc = connect(prv->socket, (struct sockaddr*)&prv->remote_un, len)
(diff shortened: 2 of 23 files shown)


drivers/block-nbd.c
```diff
@@ -942,7 +943,8 @@ tdnbd_open(td_driver_t* driver, const char* name,
 				ERROR("Failure to malloc for NBD destination");
 				return -1;
 			}
-			strcpy(prv->peer_ip, peer_ip);
+			strncpy(prv->peer_ip, peer_ip, strlen(peer_ip) + 1);
```
edwintorok, line 946: The length here should be the allocated size of the destination, otherwise you can still overflow the destination if the source is very large.

lvm/lvm-util.c
```diff
@@ -67,7 +67,8 @@ lvm_copy_name(char *dst, const char *src, size_t size)
 	if (strnlen(src, size) == size)
 		return -ENAMETOOLONG;
 
-	strcpy(dst, src);
+	strncpy(dst, src, size);
+	dst[size - 1] = '\0';
```
edwintorok, line 71: What if size is 0, does the caller guarantee it isn't?

vhd/vhd-index.c
```diff
@@ -413,9 +413,12 @@ vhd_index_add_bat(vhdi_name_t *name,
...
 	bat.vhd_block_size = vhd_block_size;
 
-	strcpy(bat.vhd_path, name->vhd);
-	strcpy(bat.index_path, name->index);
-	strcpy(bat.file_table_path, name->files);
+	strncpy(bat.vhd_path, name->vhd, sizeof(bat.vhd_path));
+	bat.vhd_path[sizeof(bat.vhd_path) - 1] = '\0';
+	strncpy(bat.index_path, name->index, sizeof(bat.index_path));
+	bat.index_path[sizeof(bat.index_path) - 1] = '\0';
+	strncpy(bat.file_table_path, name->files, sizeof(bat.file_table_path));
```
edwintorok, line 420: This is repeated so many times, would be good if there was a function that guaranteed it always correctly null terminates a string copy. I think `strlcpy` is the preferred function for that (like strncpy but always null terminates).
Although https://en.wikibooks.org/wiki/C_Programming/C_Reference/nonstandard/strlcpy#Criticism it is not part of Glibc on purpose. That mempcpy suggested there seems like a good alternative, but might be useful to wrap it in a macro so we don't need to type it out N times.

vhd/lib/libvhd-index.c
```diff
@@ -682,13 +682,14 @@ vhdi_append_block(vhdi_context_t *ctx, vhdi_block_t *block, uint32_t *sector)
...
 static int
-vhdi_copy_path_to(vhdi_path_t *path, const char *src, const char *dest)
+vhdi_copy_path_to(vhdi_path_t *path, const char *src, const char *dest, size_t dest_size)
 {
 	int len, err;
 	char *file, *relative_path, copy[VHD_MAX_NAME_LEN];
 	char *absolute_path, __absolute_path[PATH_MAX];
 
-	strcpy(copy, dest);
+	strncpy(copy, dest, dest_size);
```
edwintorok, line 691: `dest` is a source here not a destination, so the size here should be `sizeof(copy)`, otherwise a very large destsize could still cause a stack overwrite.

drivers/block-nbd.c
MarkSymsCtx: Dest is `malloc'd`  to that size 2 lines above and the result checked, so we wither fail to allocate or have enough space.

lvm/lvm-util.c
MarkSymsCtx: Even if it were the check at line 67 would have returned `-ENAMETOOLONG`

drivers/block-nbd.c
rosslagerwall: `strcpy` is guaranteed to be correct here because it is allocated to the correct size so use of `strncpy` seems unnecessary here.

In any case, it would be preferable to replace the malloc+cpy with `strdup()`.

vhd/lib/libvhd.c
```diff
@@ -2876,7 +2876,7 @@ vhd_initialize_footer(vhd_context_t *ctx, int type, uint64_t size)
 	ctx->footer.type         = type;
 	ctx->footer.saved        = 0;
 	ctx->footer.data_offset  = 0xFFFFFFFFFFFFFFFFULL;
-	strcpy(ctx->footer.crtr_app, "tap");
+	strncpy(ctx->footer.crtr_app, "tap", sizeof(ctx->footer.crtr_app));
```
rosslagerwall, line 2879: Why is this necessary for Coverity here? It is trivial to see that an overrun can not happen.

vhd/vhd-index.c
rosslagerwall: Yeah, I was going to suggest defining a `strlcpy` utility function as well.

vhd/lib/vhd-util-scan.c
```diff
@@ -504,11 +504,11 @@ vhd_util_scan_extract_volume_name(char *dst, const char *src)
 	c = strrchr(copy, '/');
 	if (c == name) {
 		/* unrecognized format */
-		strcpy(dst, src);
+		strncpy(dst, src, size);
```
rosslagerwall, line 507: Shouldn't these two instances of `strncpy` also be nul-terminated?

vhd/vhd-index.c
MarkSymsCtx: I have this in a follow-up change which I'm currently testing.

control/tap-ctl-unpause.c
```diff
@@ -54,8 +55,8 @@ tap_ctl_unpause(const int id, const int minor, const char *params, int flags,
 	message.u.params.flags = flags;
 
 	if (params)
-		strncpy(message.u.params.path, params,
-				sizeof(message.u.params.path) - 1);
+		safe_strncpy(message.u.params.path, params,
+			     sizeof(message.u.params.path) - 1);
```
rosslagerwall, line 59: The previous code was safe as it left one byte at the end for the nul terminator and previously set the entire struct to 0.

In any case, this should now be `sizeof(message.u.params.path)`, not `sizeof(message.u.params.path) - 1`.

drivers/block-nbd.c
```diff
@@ -199,9 +200,8 @@ tdnbd_stash_passed_fd(int fd, char *msg, void *data)
 		close(passed_fds[free_index].fd);
 
 	passed_fds[free_index].fd = fd;
-	strncpy(passed_fds[free_index].id, msg,
-		sizeof(passed_fds[free_index].id) - 1);
-	passed_fds[free_index].id[sizeof(passed_fds[free_index].id) - 1] = '\0';
+	safe_strncpy(passed_fds[free_index].id, msg,
+		     sizeof(passed_fds[free_index].id) - 1);
```
rosslagerwall, line 204: Should be `sizeof(passed_fds[free_index].id)`, not `sizeof(passed_fds[free_index].id) - 1`.

rosslagerwall (approved):

control/tap-ctl-unpause.c
MarkSymsCtx: That's what I get for searching and replacing :(

control/tap-ctl-unpause.c
```diff
@@ -54,8 +55,8 @@ tap_ctl_unpause(const int id, const int minor, const char *params, int flags,
 	message.u.params.flags = flags;
 
 	if (params)
-		strncpy(message.u.params.path, params,
-				sizeof(message.u.params.path) - 1);
+		safe_strncpy(message.u.params.path, params,
+			     sizeof(message.u.params.path));
 	if (secondary) {
 		err = snprintf(message.u.params.secondary,
 				sizeof(message.u.params.secondary) - 1, "%s",
```
rosslagerwall, line 62: Not related to this PR, but there's an off-by-one here. If the str to write is `sizeof(message.u.params.secondary) - 1` chars long (excluding terminator), then snprintf will truncate it to `sizeof(message.u.params.secondary) - 2` chars long (+ a nul terminator) and return `sizeof(message.u.params.secondary) - 1`. The return value is not caught by the subsequent overflow check.

rosslagerwall (approved):

control/tap-ctl-unpause.c
MarkSymsCtx: Good catch, might as well fix that as well while we're there.

rosslagerwall (approved):

TimSmithCtx (approved):

PR #360, opened 2022-05-16
CA-366761: fix off by one in tapdisk_disktype_parse_params

MarkSymsCtx:

### DIFF
--- a/drivers/tapdisk-disktype.c
+++ b/drivers/tapdisk-disktype.c
@@ -248,7 +248,8 @@ tapdisk_disktype_parse_params(const char *params, const char **_path)
 		return -ENAMETOOLONG;
 
 	memset(name, 0, sizeof(name));
-	safe_strncpy(name, params, len);
+	/* Copy just the first part of the params substring to name */
+	strncpy(name, params, len);
 
 	type = tapdisk_disktype_find(name);
 


drivers/tapdisk-disktype.c
```diff
@@ -248,7 +248,7 @@ tapdisk_disktype_parse_params(const char *params, const char **_path)
 		return -ENAMETOOLONG;
 
 	memset(name, 0, sizeof(name));
-	safe_strncpy(name, params, len);
+	safe_strncpy(name, params, len + 1);
```
edwintorok, line 251: Shouldn't `len` here denote the size of the destination, i.e. `sizeof(name)` instead of the size of the source `params`? (provided that the source IS null terminated)

rosslagerwall: No. In this case we want to copy precisely `len` characters, a subset of the full `params` string.

It is perhaps a bit confusing using `safe_strncpy()`  in this way...
(Also, the memset is now unnecessary.)

rosslagerwall (dismissed):

drivers/tapdisk-disktype.c
MarkSymsCtx: yes, I think in hindsight this is actually a case where it would be better to reinstate the original strncpy of the substring.

TimSmithCtx (dismissed):

rosslagerwall (approved):

TimSmithCtx (approved):

PR #364, opened 2022-12-16
CP-35551: Remove dependency on kernel blktap2 driver

MarkSymsCtx:


drivers/tapdisk-image.c
```diff
@@ -287,6 +289,52 @@ tapdisk_image_close_chain(struct list_head *list)
...
+find_parent_nbd_path(int prt_devnum, char **prt_nbd_path)
+{
+	int err;
+	char *pattern;
+	glob_t glbuf = { 0 };
+
+	/* This is glob pattern not a regex */
+	err = asprintf(&pattern, "%s/nbd[0-9]*.%d", BLKTAP2_CONTROL_DIR, prt_devnum);
+	if (err == -1) {
+		errno = ENOMEM;
```
rosslagerwall, line 302: I think `asprintf` should set `errno` correctly so you can just use that.

drivers/tapdisk-image.c
```diff
@@ -287,6 +289,52 @@ tapdisk_image_close_chain(struct list_head *list)
...
+		return err;
+	}
+
+	err = glob(pattern, 0, NULL, &glbuf);
+	switch (err) {
+	case GLOB_NOMATCH:
+		EPRINTF("%s, failed to find parent NBD path", pattern);
```
rosslagerwall, line 309: What's the return value of this function? Above it returns -1 and sets errno, here it returns a value from the `GLOB_*` namespace, and below it returns -errno.

rosslagerwall (commented):
Looks fine to me except for some minor complaints about error handling.

control/tap-ctl-free.c
```diff
@@ -37,28 +37,59 @@
...
+		EPRINTF("Failed to open marker file %s, %d, err=%d\n",
+			path, minor, errno);
+		goto out;
+	}
+
+	err = flock(mfd, LOCK_EX | LOCK_NB);
+	if (err == -1) {
+		EPRINTF("Unable to lock marker file %s, err = %d\n",
+			path, errno);
+		err = -errno;
```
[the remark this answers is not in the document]
MarkSymsCtx: thanks, yes, that needs looking at and making consistent.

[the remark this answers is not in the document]
MarkSymsCtx: The comment in tap-ctl.c implies that it expects (but states that it can't be assumed :() that errors should be -ve.

control/tap-ctl-allocate.c
```diff
@@ -116,19 +110,64 @@ tap_ctl_check_environment(void)
...
-	/* TO-DO: get this from a file based resource */
-	*minor = 1;
-	return 0;
+	f = open(BLKTAP2_NP_RUN_DIR, O_RDONLY);
+	if (f == -1) {
+		EPRINTF("Failed to open runtime directory %d\n", errno);
+		return errno;
+	}
+
+	flock(f, LOCK_EX);
```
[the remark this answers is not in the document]
rosslagerwall: Well, EINTR can and should be retried.

ENOLCK is probably quite unlikely, but if it happens and you continue you'd end up in a situation where two processes could be in the critical section at the same time.

jandryuk:
Hi.  Nice work on removing the blktap kernel module.  Thank you!

I'm trying to modify libxl so it writes xenstores entries to backend/vbd3 so tapback can pick them up.  /etc/xen/scripts/block-tap uses tap-ctl to create and remove the tapdisks.

One issue is that libxl waits for XenbusStateInitWait before launching the hotplug script.  tapback doesn't move to InitWait without hotplug-status, but the hotplug scripts write hotplug-status.  libxl then times out waiting for InitWait.  I am using https://github.com/xapi-project/blktap/commit/96ffa61e24d1e67ce7a3189fea27f480f18e8226 to move tapback to InitWait and things seem to work with it.  It seems like other code in tapback is waiting for hotplug-status connected, so moving to InitWait earlier should be okay for XAPI's use as well as libxl.

With this branch merged into master, I get a "No such file or directory" error when running `tap-ctl destroy`:
```
# tap-ctl list 
  286214    0    0        vhd /home/user/test.vhd
# tap-ctl destroy -p 286214 -m 0
No such file or directory
# tap-ctl list 
       -    0    -          - -
# tap-ctl free -m 0
# tap-ctl list 
```

It seems that the tapdisk process exits when `tap_ctl_close()` runs.  The subsequent `tap_ctl_detach()` then cannot find the pid.  I haven't seen this before.  Is this expected?

With tapback, there is some trickiness with physical-device.  It expects to lookup devices by major:minor.  I modified the block-tap script to write_dev with the nbd path - that is needed since libxl will need that path to pass to QEMU.  The hotplug script returns 0:0 for the device major:minor which won't be correct.  I guess tapback sees physical-device first, and then physical-device-path second.  The second one is correct and maybe takes precedence?  I have only tested a single tapdisk at a time, so minor 0 is always correct.

It seems like physical-device code should be removed and only physical-device-path should be used by tapdisk?

I have more testing to do.  There were a few instances where tapdisk prints repeats of:
```
ERROR: errno -5 at vhd_complete: /home/user/fedora1.vhd: op: 3, lsec: 26763264, secs: 1, blk: 6534, blk_offset: 30170439
ERROR: errno -5 at __tapdisk_vbd_complete_td_request: req xenvbd-20-2048.1: read 0x0008 secs @ 0x019866b8 - Input/output error
ERROR: errno -5 at vhd_complete: /home/user/fedora1.vhd: op: 3, lsec: 26763264, secs: 1, blk: 6534, blk_offset: 30170439
ERROR: errno -5 at vhd_complete: /home/user/fedora1.vhd: op: 3, lsec: 26763264, secs: 1, blk: 6534, blk_offset: 30170439
...
ERROR: errno -5 at __tapdisk_vbd_request_timeout: req xenvbd-20-2048.1 timed out, retried 120 times
ERROR: errno -5 at __tapdisk_vbd_request_timeout: req xenvbd-20-2048.1 timed out, retried 120 times
```
This may have been when I had stray tapdisks hanging around.  https://github.com/xapi-project/blktap/pull/377 is help address that issue since block-tap could find the wrong minor.

jandryuk:
My libxl patches are here if anyone wants to look: https://github.com/jandryuk/xen/commits/libxl-blktap

MarkSymsCtx:
@jandryuk if you try this updated PR it should work better. I've now got it working and integrated with xapi and the storage manager code in XenServer without the kernel driver being present on the system. There are still some integration rough edges with the toolstack which need to be resolved but they all should be that side and not in blktap.

jandryuk:
In block-tap, I had to do this:
```
    dev=$(tap-ctl create -a $target)
    # returned /run/blktap-control/tapdisk/tapdisk-0
    # get pid and minor
    find_device
    # Create nbd unix path
    dev=$( printf "/run/blktap-control/nbd%ld.%d" "$pid" "$minor" )
    write_dev $dev
```
Would it make sense for `tap-ctl create` to return the NBD path `/run/blktap-control/nbd$PID.$MINOR` instead of `/run/blktap-control/tapdisk/tapdisk-$MINOR`.  `tapdisk-$MINOR` seems useful for tapdisk-internal operations, but not useful outside of tapdisk.

jandryuk:
I'm using this branch with 4 commits on top of this PR for libxl support: https://github.com/jandryuk/blktap/tree/pr-364-libxl

it includes https://github.com/xapi-project/blktap/pull/377 because that fixes an issue with `block-tap remove`.  The other commits are all documented for why they exist.  I don't know how XAPI works, so they can use a review from that angle.

jandryuk:
I was also thinking of changing to the newer (~2012!) uri syntax: https://github.com/jandryuk/xen/commit/40e2f59e2c94e7fa7bbfbf2f562eafc30502c4ec

v-thakkar:
Hi :wave: I work at Vates and came across this PR. Love that we're getting closer to removing the dependency on blktap kernel patch to use NBD instead of it. Especially after this change in XAPI: https://github.com/xapi-project/xen-api/pull/5164/commits/6433c79b2e5ffa0654d1288377b5f5788fd274fc . I was hoping to test this PR along with #377. But I was wondering @MarkSymsCtx  if you also have some branch in SM with the related changes?

We were running into [this error](https://github.com/xapi-project/sm/blob/master/drivers/blktap2.py#L272) and were wondering if that is something we can help with. But since you mentioned above that you have some storage manager code, thought to ask you first. Thanks :)

MarkSymsCtx:
> Hi 👋 I work at Vates and came across this PR. Love that we're getting closer to removing the dependency on blktap kernel patch to use NBD instead of it. Especially after this change in XAPI: [xapi-project/xen-api@6433c79](https://github.com/xapi-project/xen-api/commit/6433c79b2e5ffa0654d1288377b5f5788fd274fc) . I was hoping to test this PR along with #377. But I was wondering @MarkSymsCtx if you also have some branch in SM with the related changes?
> 
> We were running into [this error](https://github.com/xapi-project/sm/blob/master/drivers/blktap2.py#L272) and were wondering if that is something we can help with. But since you mentioned above that you have some storage manager code, thought to ask you first. Thanks :)

As I said above https://github.com/xapi-project/blktap/pull/364#issuecomment-1522022631, this is still very much a work in progress. We do have some of the SM changes but we've had to defer all of this work to the next major platform release to avoid breaking customer workloads. Taking the opportunity of the platform release to make these breaking changes.

v-thakkar:
I agree with @jandryuk. Will appreciate the details about your schedule for the next major release @MarkSymsCtx , so that we all can work collaboratively to remove the dependency on blktap patch and don't have to keep backporting it to newer kernel versions. :)
@jandryuk Thanks for the link to your commits. I'll try to test it with XAPI next week.

control/tap-ctl-allocate.c
```diff
@@ -116,19 +110,71 @@ tap_ctl_check_environment(void)
...
+			err = -errno;
+			free(path);
+			goto out;
+		}
+
+		fid = open(path, O_CREAT | O_WRONLY, 0600);
+		if (fid == -1) {
+			err = -errno;
+			EPRINTF("Failed to create ID file %s, %d\n", path, errno);
+			goto out;
```
KevinLCtx, line 160: Should `path` not be `free`ed between these 2 lines?

KevinLCtx (dismissed):

control/tap-ctl-allocate.c
MarkSymsCtx: Yes, it should, good spot.

KevinLCtx (approved):

TimSmithCtx (approved):

PR #365, opened 2023-01-05
CP-41402: use NBD to read from parent in Intellicache

MarkSymsCtx:


MarkSymsCtx:
As a precursor to a future removal of the blktap kernel device, stop using it for the internal Intellicache data reads.

drivers/tapdisk-image.c
```diff
@@ -313,14 +363,20 @@ __tapdisk_image_open_chain(int type, const char *name, int flags,
...
+		}
+		/* Don't open parent NBD in secondary mode*/
+		flags &= ~TD_OPEN_SECONDARY;
+		err = tapdisk_image_open(DISK_TYPE_NBD, prt_nbd_path,
+					 flags | TD_OPEN_RDONLY | TD_USE_NEW_NBD,
+					 encryption, &image);
 		if (err)
 			goto fail;
 
+		free(prt_nbd_path);
```
andyhhp, line 379: This is very fragile.  It only doesn't double-free because there's no `goto err` between this and leaving the function.

I'd suggest setting to NULL explicitly, and then any future rearranging of the function will remain safe.

drivers/tapdisk-image.c
```diff
@@ -287,6 +289,53 @@ tapdisk_image_close_chain(struct list_head *list)
...
+{
+	int err;
+	char *pattern;
+	glob_t glbuf = { 0 };
+
+	/* This is glob pattern not a regex */
+	err = asprintf(&pattern, "%s/nbd[0-9]*.%d", BLKTAP2_CONTROL_DIR, prt_devnum);
+	if (err == -1) {
+		errno =-errno;
+		return err;
```
edwintorok, line 303: I think you meant 'return -errno' here (or 'err = -errno' followe by 'return err' if you really want to), otherwise this is equivalent to 'return -1', but everywhere else in this function 'err' is set to '-errno' and then returned.

andyhhp (approved):

drivers/tapdisk-image.c
```diff
@@ -287,6 +289,53 @@ tapdisk_image_close_chain(struct list_head *list)
...
+		err = -EINVAL;
+		EPRINTF("%s got more than one match, %lu", pattern, glbuf.gl_pathc);
+		goto done;
+	}
+	*prt_nbd_path = strdup(glbuf.gl_pathv[0]);
+	err = 0;
+
+done:
+	if (glbuf.gl_pathv)
+		globfree(&glbuf);
```
edwintorok, line 331: The 'if' is not needed, 'globfree' should always be safe to call (also according to the manpage the struct could contain more fields than POSIX requires and it is then unspecified whether they are allocated or not).

drivers/tapdisk-image.c
```diff
@@ -287,6 +289,53 @@ tapdisk_image_close_chain(struct list_head *list)
...
+
+	if (pattern)
```
edwintorok, line 333: I think this 'if' is redundant too, free is safe to call with a NULL pointer according to the manpage.

drivers/tapdisk-image.c
```diff
@@ -334,6 +391,9 @@ __tapdisk_image_open_chain(int type, const char *name, int flags,
 	return 0;
 
 fail:
+	if (prt_nbd_path)
```
edwintorok, line 394: same comment as above, I think it is safe and simpler to just always call free.

edwintorok (commented):
in general this looks good, could only spot one bug with 'errno = -errno', the rest are some minor suggestions.

drivers/tapdisk-image.c
MarkSymsCtx: Yes, I think you're correct, or at least intended `err = - errno;` anyway.

edwintorok (approved):
Thanks, can't spot any more bugs (but can't speak for the tapdisk specific code).

MarkSymsCtx:
There's a bug in this as a result of the review changes and the tests now fails. Need to investigate further

MarkSymsCtx:
> SM was deliberately shutting shutting down the parent on the first VM stop as it was checking for any tapdisk processes having `/dev/xen/blktap-2/tapdev<minor>` open which no longer holds true. Update SM to also look for tapdisk processes having the `nbd<pid>.<minor>` socket connected for the parent's minor id.

Now fixed

TimSmithCtx (approved):

PR #367, opened 2023-01-27
CP-33338: Various updates to NBD handling

MarkSymsCtx:
With these changes in place it is possible to use the NBD server as a complete functional replacement for the kernel tapdev device.


MarkSymsCtx:
Structured Reply was required and the short-cut of making it unsupported failed `VDI.copy` as `sparse_dd` aborts the NBD connection when it is refused.

drivers/tapdisk-nbdserver.c
```diff
@@ -429,21 +430,63 @@ receive_newstyle_options(td_nbdserver_t *server, int new_fd, bool no_zeroes)
...
-		{   
+		{
 			uint16_t flags;
-			INFO("Processing NBD_OPT_EXPORT_NAME");
+			uint32_t exportnamelen;
+			uint16_t nrInfoReq;
+
+			INFO("Processing NBD_OPT_GO");
+
+			if (recv(new_fd, &exportnamelen, sizeof(exportnamelen), 0) == -1) {
```
rosslagerwall, line 445: All these recv()s should ideally be a recv_exactly_n_bytes() function to avoid issues with short reads where the other side e.g. sends 1 byte at a time or doesn't send enough bytes.

They should also handle the case where the other side disconnects (receives 0 bytes).

drivers/tapdisk-nbdserver.c
```diff
@@ -429,21 +430,63 @@ receive_newstyle_options(td_nbdserver_t *server, int new_fd, bool no_zeroes)
...
+				ret = -1;
+				goto done;
+			}
+			exportnamelen = be32toh(exportnamelen);
+
+			exportname = malloc(exportnamelen + 1);
```
rosslagerwall, line 451: While the other side is presumably trusted, it is probably worth sanity checking exportnamelen just in case.

drivers/tapdisk-nbdserver.c
MarkSymsCtx: Will give that a try in the next set of test runs.

drivers/tapdisk-nbdserver.c
MarkSymsCtx: Only moderately trusted so yes it does need input validation.

drivers/tapdisk-nbdserver.c
rosslagerwall: There's `readit()` and `sendit()` in drivers/block-nbd.c that could be converted to utility functions. Though they need to be fixed... readit() is missing a break for the error case and both of them check `errno` when the send()/recv() returns 0 which does not set errno. That means that if errno happens to be EAGAIN from some previous call it will have an endless loop.

MarkSymsCtx:
this has now been through several rounds of testing, including testing with the devtoolset 11 Leak Sanitizer. No regressions have been found when running in "normal" mode with tapdev devices and issues encountered when running with tapdev disabled have been ticketed for future work.

- Storage Migration needs to be made to work when the storage datapath is accessed over NBD
- VM export uses excessive amounts of memory when reading from NBD socket, not a leak as it is all freed at close but an unreasonably high amount of allocation.

drivers/tapdisk-nbdserver.c
```diff
@@ -1088,30 +1095,62 @@ tapdisk_nbdserver_newclient_fd(td_nbdserver_t *server, int new_fd)
...
-		return;
+		goto failreq;
 	}
 
 	vreq = &req->vreq;
 
 	memset(req, 0, sizeof(td_nbdserver_req_t));
-	/* Read the request the client has sent */
 
+	bzero(req->id, sizeof(req->id));
```
rosslagerwall, line 1124: Why do you need to zero req->id when the whole of req was zeroed in the previous line?

rosslagerwall (approved):

drivers/tapdisk-nbdserver.c
MarkSymsCtx: we don't but that was a pre-existing issue from the looks of it.

TimSmithCtx (approved):

PR #368, opened 2023-02-15
CP-33338: fix memory usage issues  in export/import tests

MarkSymsCtx:


MarkSymsCtx:
This is now passing the export/import tests which were previously failing on hosts with 16GB of RAM. Further test suites are running for full coverage.

MarkSymsCtx:
The biggest issue here was that the NBD_BLOCK_STATUS_CMD was getting a length request of `4294966784` and then attempting to allocate that as an iovec.

andyhhp (approved):

rosslagerwall (approved):

TimSmithCtx (approved):

PR #369, opened 2023-02-16
Make CBT Mocka tests work properly with devtoolset-11

MarkSymsCtx:
The devtoolset-11 compiler, when run with -O2 on the product code replaced a `malloc` + `memset` with a call to `calloc`, which then reported memory corruption errors when trying to use the `free` function from the mocka allocator library.

### DIFF
--- a/mockatests/wrappers/wrappers.c
+++ b/mockatests/wrappers/wrappers.c
@@ -45,16 +45,38 @@ static int mock_fwrite = 0;
 static int mock_vprintf = 0;
 static int mock_fseek = 0;
 
+bool should_malloc_succeed()
+{
+	return (bool) mock();
+}
+
 void *
 __wrap_malloc(size_t size)
 {
 	bool succeed = true;
 	if (mock_malloc) {
-		succeed = (bool) mock();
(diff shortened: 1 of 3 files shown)


andyhhp (approved):

TimSmithCtx (approved):

rosslagerwall (approved):

mockatests/wrappers/wrappers.c
```diff
@@ -60,6 +65,23 @@ __wrap_malloc(size_t size)
 	return NULL;
 }
 
+
+void *
+__wrap_calloc(size_t nmemb, size_t size)
+{
+	bool succeed = true;
+	fprintf(stderr, "In calloc\n");
```
edwintorok, line 73: Are you sure this doesn't call malloc/calloc itself, resulting in infinite recursion?

edwintorok: Most other 'fprintf' are commented out, perhaps this one should be too?

MarkSymsCtx: It probably should but it's cosmetic tracing

PR #370, opened 2023-02-24
Fix use after free in NBD server

MarkSymsCtx:
Found by `gcc -fanalyzer`, most others look like false positives due to use of kernel style linked list handling

### DIFF
--- a/drivers/tapdisk-nbdserver.c
+++ b/drivers/tapdisk-nbdserver.c
@@ -283,17 +283,19 @@ tapdisk_nbdserver_set_free_request(td_nbdserver_client_t *client,
 
 void
 tapdisk_nbdserver_free_request(td_nbdserver_client_t *client,
-		td_nbdserver_req_t *req)
+			       td_nbdserver_req_t *req, bool free_client_if_dead)
 {
 	tapdisk_nbdserver_set_free_request(client, req);
-	if (unlikely(client->dead && !tapdisk_nbdserver_reqs_pending(client)))
-		tapdisk_nbdserver_free_client(client);
 
 	if (unlikely(client->n_reqs_free == (client->n_reqs / 4))) {
 		/* free requests, unmask the events */
 		tapdisk_server_mask_event(client->client_event_id, 0);
 	}
 
+	if (unlikely(free_client_if_dead &&
+		     client->dead &&
+		     !tapdisk_nbdserver_reqs_pending(client)))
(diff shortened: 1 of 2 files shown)


drivers/tapdisk-nbdserver.c
```diff
@@ -1306,10 +1309,15 @@ tapdisk_nbdserver_clientcb(event_id_t id, char mode, void *data)
...
 
 fail:
+	if (vreq)
+		tapdisk_nbd_server_free_vreq(client, vreq, false);
+
 	tapdisk_nbdserver_free_client(client);
+	return;
+
 free:
 	if (vreq)
```
rosslagerwall, line 1319: I think this is dead code now after the addition of the `return` on line 1316. The only path to get here is via `TAPDISK_NBD_CMD_DISC` which doesn't ever set `vreq`.

drivers/tapdisk-nbdserver.c
```diff
@@ -283,17 +283,19 @@ tapdisk_nbdserver_set_free_request(td_nbdserver_client_t *client,
...
 	tapdisk_nbdserver_set_free_request(client, req);
-	if (unlikely(client->dead && !tapdisk_nbdserver_reqs_pending(client)))
-		tapdisk_nbdserver_free_client(client);
 
 	if (unlikely(client->n_reqs_free == (client->n_reqs / 4))) {
 		/* free requests, unmask the events */
 		tapdisk_server_mask_event(client->client_event_id, 0);
 	}
 
+	if (unlikely(free_client_if_dead &&
```
rosslagerwall, line 295: I wonder if it would be better moving this into a separate function that callers can call if they need to rather than having an awkward sounding boolean variable?

drivers/tapdisk-nbdserver.c
MarkSymsCtx: Actually, it looks like it was dead in the previous sets of changes. Just cleaning the whole thing up.

drivers/tapdisk-nbdserver.c
MarkSymsCtx: It's called, either directly or indirectly, from quite a number of places and whilst the the boolean arg is a bit icky and smelly I couldn't see making this something the callers have to do being any nicer.

rosslagerwall (approved):

edwintorok (approved):

PR #373, opened 2023-03-28
CP-37221: Optimise data management in vhd-util coalesce

MarkSymsCtx:
Reduce the amount of data read during coalesce. Especially, do not read up the chain of parents when all that is required is the allocation block from the current VHD.

### DIFF
--- a/vhd/lib/libvhd.c
+++ b/vhd/lib/libvhd.c
@@ -2041,19 +2041,14 @@ fail:
 }
 
 int
-vhd_read_block(vhd_context_t *ctx, uint32_t block, char **bufp)
+vhd_read_at(vhd_context_t *ctx, uint64_t block, uint32_t from, size_t size, char *buf)
 {
+	/*
+	 * Sub call function, assumes that dynamic disk has been checked
+	 */
 	int err;
-	void *buf;
-	size_t size;
 	uint64_t blk;
 	off64_t end, off;
-
-	buf   = NULL;
-	*bufp = NULL;
-
-	if (!vhd_type_dynamic(ctx))
-		return -EINVAL;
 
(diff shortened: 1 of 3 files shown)


vhd/lib/vhd-util-coalesce.c
```diff
@@ -100,6 +100,10 @@ vhd_util_coalesce_block(vhd_context_t *vhd, vhd_context_t *parent,
 	if (err)
 		goto done;
 
+	err = posix_memalign((void *)&buf, 4096, vhd->header.block_size);
+	if (err)
+		return -err;
```
andyhhp, line 105: Given the `goto done` immediately above, shouldn't this match ?

MarkSymsCtx: yes, it should as the order of allocation of the block and bitmap has changed so we need to free the bitmap if we fail to memalign the block.

vhd/lib/libvhd.c
```diff
@@ -2067,34 +2062,55 @@ vhd_read_block(vhd_context_t *ctx, uint32_t block, char **bufp)
...
 
+	off = off + vhd_sectors_to_bytes(from);
+
 	err  = vhd_seek(ctx, off, SEEK_SET);
 	if (err)
-		goto fail;
+		return err;
 
 	err  = vhd_read(ctx, buf, size);
+	if (err)
```
edwintorok, line 2082: This is correct, but wouldn't it be simpler to `return vhd_read(ctx, buf, size)`?

edwintorok (approved):
More cleanup opportunities discovered, but out of scope for this PR.

TimSmithCtx (approved):

PR #378, opened 2023-05-09
Add an option to never resolve parent path when vhd-util query is called

Wescoeur:
Add the ability to vhd-util to find the parent of a VHD without locally resolving or calling stat on the locator path.
A cmdline option `-u` (unresolved) is now available to query the parent path without resolution.

Use cases:
- When a special SR driver is used where the parent is not always accessible on the master (DRBD case).
- A copy of a VHD on an external device without parent
- ...

Example:
```
vhd-util query -u -n /dev/drbd/by-res/xcp-volume-1d86819f-c844-4250-8f3a-df5de9227baa/0 -p
../xcp-volume-56c6ef19-3d87-4838-b5e6-9c534bc1a0cb/0
```

### DIFF
--- a/include/libvhd.h
+++ b/include/libvhd.h
@@ -341,6 +341,7 @@ int vhd_initialize_header_parent_name(vhd_context_t *, const char *);
 int vhd_write_parent_locators(vhd_context_t *, const char *);
 int vhd_parent_locator_count(vhd_context_t *);
 int vhd_parent_locator_get(vhd_context_t *, char **);
+int vhd_parent_locator_unresolved_get(vhd_context_t *, char **);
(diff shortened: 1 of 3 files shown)


benjamreis (approved):
LGTM
@MarkSymsCtx what do you think of this change?
Would that be of interest to you?

MarkSymsCtx (approved):
It looks like quite a specialised case which may not get a lot of use but you do obviously have a use for it so it looks good to me.

TimSmithCtx (approved):

stormi:
Is something missing to reach the point of merging?

Wescoeur:
We have been using this change for over a year without problems, would it be possible to merge it on your side? Thank you! :)

PR #380, opened 2023-07-26
CA-380565: return ETIMEDOUT when -t timeout is hit in IPC select

MarkSymsCtx:


control/tap-ctl-ipc.c
```diff
@@ -61,6 +61,7 @@ tap_ctl_read_raw(int fd, void *buf, size_t size, struct timeval *timeout)
 	fd_set readfds;
 	size_t offset = 0;
 	int ret;
+	int err = 0;
 
 	while (offset < size) {
 		FD_ZERO(&readfds);
```
edwintorok, line 67: This only watches one file, the zeroing could be moved out of the loop, no need to zero all 1024 on each iteration. Alternatively `poll` could be used instead of select if you want to watch just 1 file (select might still have to traverse the bitmap to find out what to watch).

control/tap-ctl-ipc.c
```diff
@@ -74,13 +75,17 @@ tap_ctl_read_raw(int fd, void *buf, size_t size, struct timeval *timeout)
...
+			err = ETIMEDOUT;
 			break;
+		}
+
 	}
 
 	if (offset != size) {
 		EPRINTF("failure reading data %zd/%zd\n", offset, size);
-		return -EIO;
+		return err ? - err : -EIO;
```
edwintorok, line 88: Shouldn't this return the actual errno instead of EIO if there is one? 'err' could be set to 'errno' after 'select' and 'read', leaving 'EIO' just for the 'EOF encountered too soon' error.

control/tap-ctl-ipc.c
MarkSymsCtx: Probably too big a change to switch to poll, it might be reasonable in a future change but at the moment just trying to fix the misreporting of the time out so that the caller can make sensible decisions in the case of a time out.

rosslagerwall (approved):

TimSmithCtx (approved):

PR #381, opened 2023-09-05
CP-40871: return the number of sectors coalesced from vhd-util

MarkSymsCtx:


vhd/lib/vhd-util-coalesce.c
```diff
@@ -138,15 +150,27 @@ vhd_util_coalesce_block(vhd_context_t *vhd, vhd_context_t *parent,
 done:
 	free(buf);
 	free(map);
-	return err;
+	if (err < 0)
```
KevinLCtx, line 153: It would be kind of bad if some function call returns a positive error code and that gets misinterpreted as the number of sectors coalesced. Line 105/116 after `posix_memalign` shows that we expect positive error codes to be a possibility. We could check for `err > 0` here and invert the result for extra safety. Especially the guard against somebody making careless changes in the future.

MarkSymsCtx: Everything should return -ve error codes, the internal vhd_xxxx calls do, so it's only returns from library calls (like posix_memalign) which need care and are in this case inverted in the error handler block.

TimSmithCtx (approved):

PR #382, opened 2023-09-06
Rationalise blktap io errors

MarkSymsCtx:
Rationalise the IO error handling in the block handlers. In a scenario where the NFS server connection has dropped the retires can add up to an extremely large amount of time being spent before the IO requests are failed back to the VM.

### DIFF
--- a/drivers/block-vhd.c
+++ b/drivers/block-vhd.c
@@ -1905,7 +1905,7 @@ schedule_bitmap_read(struct vhd_state *s, uint32_t blk)
 	return 0;
 }
 
-static void
+static bool
 schedule_bitmap_write(struct vhd_state *s, uint32_t blk)
 {
 	uint64_t offset;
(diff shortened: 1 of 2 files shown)


KevinLCtx (dismissed):

drivers/block-vhd.c
```diff
@@ -1905,25 +1905,31 @@ schedule_bitmap_read(struct vhd_state *s, uint32_t blk)
...
 
 	bm     = get_bitmap(s, blk);
+
 	offset = bat_entry(s, blk);
 
 	ASSERT(vhd_type_dynamic(&s->vhd));
 	ASSERT(bm && bitmap_valid(bm) &&
 	       !test_vhd_flag(bm->status, VHD_FLAG_BM_WRITE_PENDING));
 
+	if (memcmp(bm->map, bm->shadow, 512) == 0) {
```
rosslagerwall, line 1923: This `512` should be calculated like `map_size` in `finish_bitmap_transaction`.

rosslagerwall (commented):
Can you add the relevant ticket IDs to the commits?

TimSmithCtx (approved):

PR #384, opened 2023-11-23
CP-46757: advertise feature-flush-cache

MarkSymsCtx:
Windows guests do not request BLKIF_OP_WRITE_BARRIER and only support BLKIF_OP_FLUSH_DISKCACHE, so advertise and support that to allow for explicit cache and data integrity management.


drivers/td-req.c
```diff
@@ -770,6 +771,8 @@ tapdisk_xenblkif_make_vbd_request(struct td_xenblkif * const blkif,
         tapreq->prot = PROT_READ;
         vreq->op = TD_OP_WRITE;
         break;
+    case BLKIF_OP_FLUSH_DISKCACHE:
```
edwintorok, line 774: Should we have a sanity check here that number of segments is 0? Also should we have some stats counters so we can tell whether the feature is being used or not?

edwintorok, line 774: Should we also forward this to the underlying storage? The documentation at [https://docs.xenserver.com/en-us/xenserver/8/developer/ddk implies](https://github.com/xen-project/xen/blob/master/xen/include/public/io/blkif.h#L551-L554) implies that this is stronger than a barrier at the tapdisk level: we should also seek some confirmation from the actual underlying storage (O_DIRECT is not enough for that, it can in turn be still cached in volatile storage controller cache).

MarkSymsCtx: There is, see line 788-791 below.

MarkSymsCtx: There's not really any mechanism to do that, there is no IO associated with a BLKIF_OP_FLUSH_DISKCACHE so nothing that e.g. an FUA could be attached to.

edwintorok: Isn't that test the other way around? if the number of segments is zero check that it is one of the flush operations.
But if you get a flush disk cache with a non-zero segments, what happens then?

MarkSymsCtx: You can't, it's a dataless operation.

MarkSymsCtx: This needs to move, a new change is being updated and tested

MarkSymsCtx: Libaio doesn't support sending an FUA or equivalent. I guess you could do something in the completion path to call `SG_IO ioctl` or `fsync` depending on whether it's block or file based but that's probably a bit complicated.

PR #394, opened 2024-01-24
libxl blktap support

jandryuk:
These two patches are the minimal needed for integration with libxl.

The InitWait handling needs a little modification to set it earlier.  libxl needs to see InitWait to run hotplug scripts which will write hotplug-status = "connected".  Setting InitWait earlier won't allow transitioning to Connected any earlier, so I *think* it should be okay with XAPI.  Howver, it has *not* been tested under XAPI.

The deleting of the xenstore backend is no longer performed.  libxl expects to find the backend to run the hotplug scripts to perform cleanup.  Later on, libxl will delete the entire domain sub-tree, so xenstore will get cleaned up that way.  Again, untested under XAPI.  Maybe XAPI does similar sub-tree cleanup so this isn't necessary?

If this isn't acceptable as-is, would a command line option & global flag for a "libxl mode" be acceptable to change behavior?

### DIFF
--- a/tapback/backend.c
+++ b/tapback/backend.c
@@ -259,6 +259,15 @@ tapback_backend_create_device(backend_t *backend,
         goto out;
     }
 
+    /* After tapback has written it's capabilities to XenStore, switch to
+     * InitWait. */
+    err = xenbus_switch_state(device, XenbusStateInitWait);
+    if (unlikely(err)) {
+        WARN(device, "failed to switch to XenbusStateInitWait: %s\n",
+             strerror(-err));
+        goto out;
+    }
+
 out:
     if (err) {
         WARN(NULL, "%s: error creating device: %s\n", name, strerror(-err));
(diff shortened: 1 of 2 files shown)


MarkSymsCtx:
I'll run this through our test systems and check nothing goes bang,

MarkSymsCtx:
OK, as this stands right now it causes Xenopsd to fail with timeouts so it can't be immediately merged in this form. I'll get somebody to look at the xenopsd failures and see if that can be made to work compatibly.

jandryuk:
@MarkSymsCtx Thanks so much for testing it - sorry it didn't work...

MarkSymsCtx:
@jandryuk no problems, we'll look into whether we can get xenopsd to function compatibly with this. It looked to be timing out waiting on a state change, which possibly means that it's missed that the transition has already occurred.

jandryuk:
This [branch](https://github.com/jandryuk/blktap/commits/libxl-blktap-flag/) puts the changes from this PR behind a new `-l`/`--libxl` flag https://github.com/jandryuk/blktap/commit/63f3883abe1f94a85428e17e98fe0dbe15606b0b

PR #398, opened 2024-03-26
CA-390490: allocate vhd request before locking bat

MarkSymsCtx:
In the possible event that there were no available vhd_request available it was possible that the operation would be failed having marked the BAT as locked with no attempt to unwind that lock. As it happens unwinding the lock could potentially be difficult as it is permitted for multiple requests which are updating the same newly created allocation block to be submitted so, just because the VHD_FLAG_REQ_UPDATE_BAT flag is set it is not certain that this request was the first to lock. To simplify this, first acquire the vhd_request which has simple cleanup if it fails, then acquire the BAT lock and release the vhd_request if this fails.

### DIFF
--- a/drivers/block-vhd.c
+++ b/drivers/block-vhd.c
@@ -1780,38 +1780,9 @@ schedule_data_write(struct vhd_state *s, td_request_t treq, vhd_flag_t flags)
 	uint64_t offset;
 	uint32_t blk = 0, sec = 0;
 	struct vhd_bitmap  *bm = NULL;
-	struct vhd_request *req;
+	struct vhd_request *req = NULL;
 	char *crypto_buf = NULL;
 
-	if (s->vhd.footer.type == HD_TYPE_FIXED) {
-		offset = vhd_sectors_to_bytes(treq.sec);
-		goto make_request;
-	}
-
-	blk    = treq.sec / s->spb;
-	ASSERT(blk < s->bat.bat.entries);
-	sec    = treq.sec % s->spb;
-	offset = bat_entry(s, blk);
-
-	if (test_vhd_flag(flags, VHD_FLAG_REQ_UPDATE_BAT)) {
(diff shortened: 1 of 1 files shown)


edwintorok (approved):
This is probably best reviewed locally with `git` configured to show code moves in a different highlight color, resulting in a much smaller diff:
```
[diff]
  colorMoved = dimmed_zebra
```

With that setting I see the following changes:
* 'goto make_request' from HD_TYPE_FIXED is dropped, and instead turned into an 'else' branch. This is fine because the code got moved after make_request, and it shouldn't have used goto in the first place
* `if err return err` from UPDATE_BAT is replaced with a `goto fail`. This is expected as we have more to deallocate
* the `make_request` goto label is dropped
* a typo is fixed in the comment above the ASSERT >= 1536 line
* alloc_vhd_request failure path is replaced with a goto fail and conditional 'free(crypto_buf)' is moved there, this is expected
* a 'free_vhd_request' is introduced as expected because we can now fail after having allocated a request

edwintorok:
Not part of this PR change, but why does a 'posix_memalign' failure get turned into EBUSY instead of ENOMEM? I understand why 'alloc_vhd_request' would be EBUSY (we have a fixed number of slots, and a failure to allocate one doesn't mean we are OOM, it just means we have too many in-flight requests), but a posix_memalign failure really means you're OOM...

MarkSymsCtx:
> Not part of this PR change, but why does a 'posix_memalign' failure get turned into EBUSY instead of ENOMEM? I understand why 'alloc_vhd_request' would be EBUSY (we have a fixed number of slots, and a failure to allocate one doesn't mean we are OOM, it just means we have too many in-flight requests), but a posix_memalign failure really means you're OOM...

Indeed, a problem for another day, probably somebody being overly hopeful that going EBUSY and retying might see the memory having been release by another request.

KevinLCtx (approved):

PR #402, opened 2024-07-25
CA-391882: fix nbd connection fd leak bug

chunjiez:
When nbd server receives disconnect request from client, it frees
  client data structure but it does not close the connection, then
  it sends VDI size to client, the send most likely returns err as
  client fd has been closed. However, in some case like half close,
  send returns successfully, and then server will always keep this
  opened fd in memory and never close it, finally, the number of
  opened fd exceeds ulimit settings.


MarkSymsCtx (approved):
Pending results of tests but the change looks OK so long as we don't encounter any bugs in it.

chunjiez:
@MarkSymsCtx    See suite test 202001 (Dev Run). I add the commit as a patch file into blktap.spec and generate my private build.

TimSmithCtx (approved):

PR #404, opened 2024-08-08
CA-396659: rationalise bitops and fix overflow issues

MarkSymsCtx:
Has completed a BST suite but pending being re-run with UBSAN which found the issues originally


edwintorok (dismissed):
The changes look good, and it is nice that this now has some actual tests too.

rdn32:
The implementation of new_set_bit in vhd-update.c was the same as the old implementation of set_bit (but with its arguments in a different order) so surely it would suffer from the same problem. Should that be changed too?

rdn32 (approved):

MarkSymsCtx:
> The implementation of new_set_bit in vhd-update.c was the same as the old implementation of set_bit (but with its arguments in a different order) so surely it would suffer from the same problem. Should that be changed too?

Actually the answer is that this is an ancient legacy bit of technical debt which is no longer required, so dropped. Good spot though :)

vhd/lib/libvhd.c
```diff
@@ -744,7 +726,7 @@ vhd_bitmap_test(vhd_context_t *ctx, char *map, uint32_t block)
 {
 	if (vhd_creator_tapdisk(ctx) &&
```
MarkSymsCtx, line 727: These should probably now be fatal errors, if there are any VHDs still out there which were created with the broken 0.1 version prior to May 2008 then they're long past due being removed.

MarkSymsCtx:
Tests completed with UBSAN enabled and all passed. Follow-up change to make the unit tests work with UBSAN.

vhd/lib/libvhd.c
MarkSymsCtx: now made to abort

include/libvhd.h
```diff
@@ -165,21 +163,21 @@ struct vhd_context {
...
-	return ((addr[nr >> 3] << (nr & 7)) & BIT_MASK) != 0;
+	return (((const uint8_t *)addr)[nr >> 3] >> (nr & 7)) & 1;
 }
 
 static inline void
-set_bit (volatile char *addr, int nr)
+set_bit (void *addr, uint32_t nr)
 {
-	addr[nr >> 3] |= (BIT_MASK >> (nr & 7));
+	((uint8_t *)addr)[nr >> 3] |= (1 << (nr & 7));
```
andyhhp, line 174: These two lines are not equivalent.  Consider `nr` being a multiple of 8, so `nr & 7` is 0.  In the old code you tested the top bit of the byte, and in the new code you're testing the bottom bit.

Overall, you've done a bit-reverse on the byte as it's stored in the bitmap.

AFAICT, the `old_*_bit()` were reasonably sane (on a little endian system at least), whereas the new `*_bit()` are insane on both little and bit endian systems.

andyhhp: Yeah, spec says it's byte granular bit-reversed.

https://github.com/libyal/libvhdi/blob/6c102a20ad8d0c93330bc6a97beb97718befd26b/documentation/Virtual%20Hard%20Disk%20(VHD)%20image%20format.asciidoc?plain=1#L354-L355

This deserves a comment explaining the format of the bitmap.

mockatests/vhd/test-bitops.c
```diff
@@ -0,0 +1,65 @@
...
+        set_bit(map, bit);
+        assert_true(test_bit(map, bit));
+
+        /* Clear it and check */
+        clear_bit(map, bit);
+        assert_false(test_bit(map, bit));
+    }
+
+    free(map);
+}
```
andyhhp, line 65: This test doesn't really do anything useful, besides confirm that all 3 helpers implement the same algorithm.

What you really want is something like (untested):
```c
uint8_t map[2] = { 0x80, 0x01 };

assert_true(test_bit(map, 0));
assert_false(test_bit(map, 1));

assert_false(test_bit(map, 14));
assert_true(test_bit(map, 15));

set_bit(map, 3);
assert_true(map[0] == 0x90);

clear_bit(map, 0);
assert_true(map[0] == 0x10);
```
or so, which also checks the helper algorithms against the raw format.

include/libvhd.h
```diff
@@ -165,21 +163,21 @@ struct vhd_context {
 };
 
 static inline int
-test_bit (volatile char *addr, int nr)
+test_bit (const void *addr, uint32_t nr)
 {
-	return ((addr[nr >> 3] << (nr & 7)) & BIT_MASK) != 0;
+	return (((const uint8_t *)addr)[nr >> 3] >> (nr & 7)) & 1;
```
andyhhp, line 168: If you can bare to swap `int` for `bool`, you can drop the final `& 1`.

vhd/lib/libvhd.c
```diff
@@ -753,8 +737,10 @@ void
 vhd_bitmap_set(vhd_context_t *ctx, char *map, uint32_t block)
 {
 	if (vhd_creator_tapdisk(ctx) &&
-	    ctx->footer.crtr_ver == 0x00000001)
-		return old_set_bit(map, block);
+	    ctx->footer.crtr_ver == 0x00000001) {
+		VHDLOG("error unsupported version 0.1 VHD");
+		abort();
+	}
```
andyhhp, line 743: IMO this ought to be in vhd_open() and to return an API error rather than coredumping.

As a bonus that removes logic from what is clearly a semi-fast (if not fully fast) path.

mockatests/vhd/test-bitops.c
MarkSymsCtx: Both I think, the existing test exercises the full range and therefore ensures we don't hit the UBSAN overflow errors and then some "static" tests to prove the algorithm.

include/libvhd.h
MarkSymsCtx: yep, I had actually intended to do just that and forgot

rdn32 (approved):

MarkSymsCtx:
This is being reworked based on comments above

MarkSymsCtx:
Testing uncovered further undefined behaviour in `scheduler.c:scheduler_get_event_uuid` which are also addressed by handling overflow conditions explicitly.

MarkSymsCtx:
Builds with devtoolset-11 (and related sanitiser packages) and these CFLAG controls
```
-fsanitize=undefined -fsanitize-undefined-trap-on-error -fno-sanitize=shift-base
```
now pass the build time unit tests and standard stability test suites.

rdn32 (approved):

andyhhp (approved):
This does now look correct, and an improvement over the status quo.

However, I think it would be wise to take `test_bit()` et al out of libvhd.h and inline them into `vhd_bitmap_*()`.  block-log.c has its own, different, bitmap helpers (sensibly little endian) which will explode in fun ways if libvhd.h is ever included.

PR #405, opened 2024-08-16
CP-50952 Autodetect NBD server protocol version on client connect

TimSmithCtx:
Rather than using a flag on the driver state to determine what NBD protocol to use, simply detect the NBD protocol by version by looking at the second NBD magic field transmitted by the server.

The TD_DRIVER_NEW_NBD flag thus becomes redundant and can be removed, which in turn renders the TD_USE_NEW_NBD flag unnecessary.

We do not need two identical defines for the same concept in the same file. Retain the separate NBD header file for now even though it gives us multuple defines of the old-style handshake data.


drivers/block-nbd.c
```diff
@@ -615,7 +624,7 @@ tdnbd_nbd_negotiate_old(struct tdnbd_data *prv, td_driver_t *driver)
 	 *
```
MarkSymsCtx, line 624: Should we have a similar comment in `tdnbd_nbd_negotiate_new` below?

drivers/block-nbd.c
```diff
@@ -718,92 +675,96 @@ tdnbd_nbd_negotiate_old(struct tdnbd_data *prv, td_driver_t *driver)
...
 tdnbd_nbd_negotiate(struct tdnbd_data *prv, td_driver_t *driver)
 {
-	if(CLIENT_USE_OLD_HANDSHAKE && !td_flag_test(driver->state, TD_DRIVER_NEW_NBD)) {
+	int rc;
+	uint64_t magic;
+	int sock = prv->socket;
+
+	/* Read the NBD opening magic number, which is the same for all protocol
+	 * versions */
+    rc = tdnbd_wait_recv(sock, &magic, sizeof(magic), 0);
```
MarkSymsCtx, line 731: mixed tabs and spaces? and 747. (Yes, I know, we really need to run uncrustify on this code)

MarkSymsCtx (changes requested):
Looks generally good, couple of minor comments.

drivers/block-nbd.c
TimSmithCtx: Perhaps, and we would have if there had been a comment there already that could be modified.

I'll go and make one...

MarkSymsCtx (dismissed):
LGTM

drivers/block-nbd.c
```diff
@@ -683,33 +650,23 @@ tdnbd_nbd_negotiate_old(struct tdnbd_data *prv, td_driver_t *driver)
...
-		}
-
-		rc = recv(sock, buffer, padbytes, 0);
+		rc = tdnbd_wait_recv(sock, buffer, padbytes, 0);
 		if (rc < 0) {
-			ERROR("Bad read in negotiation(5) (%d)\n", rc);
-			close(sock);
-			return -1;
+			ERROR("Error in nbd_negotiate: %s", strerror(-rc));
+			// goto errout;
```
rdn32, line 669: Should the commented-out code be restored?

drivers/block-nbd.c
```diff
@@ -554,68 +563,68 @@ negotiate_client_newstyle_options(int sock, td_driver_t *driver)
...
-		close(sock);
-		return -1;
+		goto errout;
 	}
 
 	/* Collect the results in the handshake finished */
-	struct nbd_export_name_option_reply handshake_finish;
-	static const size_t NO_ZERO_HANDSHAKE_FINISH_SIZE = 10;
-	rc = recv_fully_or_fail(sock, &handshake_finish, NO_ZERO_HANDSHAKE_FINISH_SIZE);
+	rc = tdnbd_wait_recv(sock, &handshake_finish, NO_ZERO_HANDSHAKE_FINISH_SIZE, 0);
```
rdn32, line 585: The code used to use recv_fully_or_fail, so it looks like it's assuming that if recv doesn't fail then it will always read the full amount. Elsewhere the same assumption is being made. Is there something at a more global level that's making sure this is OK?

drivers/block-nbd.c
TimSmithCtx: Yes. Looks like a "comment this line" hotkey combination got activated without me noticing...

drivers/block-nbd.c
TimSmithCtx: I would like to draw your attention to a few lines below, where
```
rc = fcntl(sock, F_SETFL, O_NONBLOCK);
```
happens. The socket is still blocking at this point, so if we get to the recv(), it _will_ receive NO_ZERO_HANDSHAKE_FINISH_SIZE octets. ```tnbd_wait_recv()``` ensures the socket has become readable before it attempts to read, which ensures that the ```recv()``` will not block forever.

At this point anything other than a full read indicates that Something Bad happened to the peer.

MarkSymsCtx (approved):

rdn32 (approved):

PR #407, opened 2024-08-30
CA-397082: free nbd client connection appropriately

chunjiez:
There are 2 different kinds of nbd clients. One is nbd block device in kernel, the other is SXM sparse_dd data copy. And these 2 clients perform differently during disconnection and connection establishment.

In case 1, client is nbd block device, server side does not close socket fd on receiving NBD_CMD_CLOSE request, see tapdisk_nbdserver_free_client function. It might cause fd leak issue.

In case 2, client is SXM, during SXM, there are some tap-ctl pause and unpause operations, client sends NBD_CMD_CLOSE request to server and waits for server reconnecting to client on same socket fd, either client or server does not close the socket fd.

https://github.com/NetworkBlockDevice/nbd/blob/master/doc/proto.md#terminating-the-transmission-phase, see NBD protocol, it looks like tapdisk NBD implementation is not compatible with protocol.

In order to keep the code patch small, add nanosleep inside tapdisk_nbdserver_free_client, and the code change meets requirements of both above 2 clients.

Test,
xenrt job, 4085026, storage-next private build, OK
xenrt job, 4085027, rawhide private build, OK
xenrt job, 4089119, storage-next private build with latest blktap code, OK

### DIFF
--- a/drivers/tapdisk-nbdserver.c
+++ b/drivers/tapdisk-nbdserver.c
@@ -823,8 +823,6 @@ tapdisk_nbdserver_free_client(td_nbdserver_client_t *client)
 	if (likely(!tapdisk_nbdserver_reqs_pending(client))) {
 		list_del(&client->clientlist);
 		tapdisk_nbdserver_reqs_free(client);
-		if (client->client_fd > 0)
-			close(client->client_fd);
 		free(client);
 	} else
 		client->dead = true;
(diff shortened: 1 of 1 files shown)


chunjiez:
@MarkSymsCtx @TimSmithCtx  Could you please review this PR?

chunjiez:
Storage fellows, could you please post some comments on this PR, or give +1 if you think the code change is good?

MarkSymsCtx:
@chunjiez we're not accepting this change. @TimSmithCtx is fixing this properly, and reverting the original change that caused this problem in the first place.

PR #422, opened 2025-04-07
CP-54256: log errors when reporting EOPNOTSUP

MarkSymsCtx:
In cases where a request cannot be dispatched because a driver doesn't implement the target function log the error before returing the error code.

Additionally log the image name when processing a failure in td_complete_request.


MarkSymsCtx:
All other locations where EOPNOTSUPP is reported already have error level logging.

KevinLCtx (approved):

PR #430, opened 2025-08-08
CA-400404: remove low memory mode

MarkSymsCtx:
Low memory mode relied on cgroups v1 memory.pressure.level, now that we are cgroupss v2 this no longer exists. Pressure Stall Informationnn does exist and would generate events if the process stalls. This basically really only happens when the process is "thrashing", i.e. all memory and swap is exhausted. The intent of low memory mode was to prempt the system getting to this state.

Since this functionality was originally implemented servers have got bigger and more RAM is readily available. The outcome of low memory mode is to throttle the I/O of the VM(s) into essentially queue depth 1 linear I/O which will be significantly impacting to workloads. Increasing the size of Dom0 memory to accommodate the expected workloads will give a better overall experience. Modern VMs now also have far more per-vm back-end services running in Dom0, e.g. varstored for UEFI VMs, vTPM processes, etc.


drivers/tapdisk-metrics-stats.h
```diff
@@ -34,8 +34,6 @@
 
 #include <stdint.h>
 
-#define BT3_LOW_MEMORY_MODE 0x0000000000000001
```
rosslagerwall, line 37: Is there also code to be removed where the stats are parsed?

MarkSymsCtx: put the flags back

rosslagerwall (approved):

LunfanZhang (approved):

TimSmithCtx (approved):

PR #435, opened 2026-01-09
CA-395093: sync between tapback slave process and xenopsd

chunjiez:
In some corner case, physical-device-path xenstore watch event is fired before slave tapback process ready to process xenstore watch event, thus, slave tapback process would miss xenstore watch event, then blktap io datapath fails to establish.

In xenopsd side, the vbd-script waits for tapback slave process ready by checking /var/run/tapback.<N>.statefile, if the file is present and file contains "ping" string, then vbd-script updates the file, writes "pong" to the file and continues to update xenstore, otherwise, just wait.

In tapback slave process side, once it get prepared to process xenstore watch event, it writes "ping" string to /var/run/tapback.<N>.statefile, then waits for acknowledge by checking if the file contains "pong" string, after seeing "pong" string, it removes /var/run/tapback.<N>.statefile and continues to work.

### DIFF
--- a/drivers/td-blkif.h
+++ b/drivers/td-blkif.h
@@ -174,7 +174,7 @@ struct td_xenblkif {
 
 	struct {
 		/**
-		 * Pointer to the pending barrier request.
+		 * Pointer to he pending barrier request.
 		 */
 		blkif_request_t *msg;
 
--- a/drivers/td-req.c
+++ b/drivers/td-req.c
@@ -759,17 +759,6 @@ tapdisk_xenblkif_make_vbd_request(struct td_xenblkif * const blkif,
         tapreq->prot = PROT_WRITE;
         vreq->op = TD_OP_READ;
         break;
-    case BLKIF_OP_FLUSH_DISKCACHE:
-        /*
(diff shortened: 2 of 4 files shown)


chunjiez:
The xenopsd side code update, https://github.com/xapi-project/xen-api/pull/6825

tapback/tapback.c
```diff
@@ -222,6 +222,71 @@ tapback_write_pid(const char *pidfile)
...
+        goto fail;
+    }
+
+    /* write temp file */
+    fp = fopen(tmpfile, "w");
+    if (!fp) {
+        err = errno;
+        goto fail;
+    }
+    err = fprintf(fp, "ping");
```
MarkSymsCtx: This should follow the precendent set by the PV ring handshake and use integer state values.

MarkSymsCtx (changes requested):
This all feels like it is adding complexity and risk to an already fragile part of the VM start process. I can easily see the tapback process getting stuck waiting for the response. I think we need to find a better way of achieving the aims of this pair of PRs.

